PR #99, opened 2021-07-26
Add coro::resume_order_policy for coro::event

jbaldwin:
lifo + fifo!

When setting a `coro::event` it is now possible to declare the order in which you want the waiters to be resumed.  Default is lifo for performance but fifo can be picked at a runtime perf cost.

### DIFF
--- a/inc/coro/event.hpp
+++ b/inc/coro/event.hpp
@@ -7,6 +7,16 @@
 
 namespace coro
 {
+enum class resume_order_policy
+{
+    /// Last in first out, this is the default policy and will execute the fastest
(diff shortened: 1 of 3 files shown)


niansa (changes requested):
It _might_ be better for performance to use seperate (templated) classes instead to decrease amount of branches but from what I have seen the code makes perfect sense.

jbaldwin:
Yeah I started with it being templated version to remove the branch but then it doesn't compose as well across other usages in the library since it has two concrete types then.  I might play around with it as a follow up to see how bad it cascades across everything but for now I'll get this merged so its available.

PR #101, opened 2021-07-27
attempt to reproduce example.com with ssl hang

jbaldwin:


jbaldwin:
Not fixing right now

niansa:
rip

PR #110, opened 2021-11-27
Fixes #109 fn overload on task promise::return_void bug

### LINKED ISSUE #109: coco::detail::promise<void>::return_void(): Misplaced rethrow
Greetings.

I'm looking at task.hpp here.

You're using coro::detail::promise<void>::return_void() to rethrow the exception after the end of the coroutine. That part I understand.

But according to the spec, promise::return_void() is also called automatically during co_return. From the coroutine itself. Where a rethrow makes no sense.

This is a strange double use of the function, and I think it's just a bug. Or I don't get it, which is always a possibility.
In the case of promise-of-not-void you also use return_value() function for two purposes, but that works since they are distinct by parameter overload.

Best Regards, Jørn

jbaldwin:
The return_void() overload was throwing, but this is
incorrect per the standard, it should effectively do
nothing.  This was an error on my part since I over-
loaded the non-void task<T> return_value(T) -> void
and return_value() -> T fns.  This overload isn't
possible with the void versions since the signature
is identical, thus the bug caused the compiler's coroutine
version to re-throw exceptions at the wrong time.

### DIFF
--- a/inc/coro/task.hpp
+++ b/inc/coro/task.hpp
@@ -70,7 +70,7 @@ struct promise final : public promise_base
 
     auto return_value(return_type value) -> void { m_return_value = std::move(value); }
 
-    auto return_value() const& -> const return_type&
+    auto result() const& -> const return_type&
     {
         if (m_exception_ptr)
         {
@@ -80,7 +80,7 @@ struct promise final : public promise_base
         return m_return_value;
     }
 
-    auto return_value() && -> return_type&&
+    auto result() && -> return_type&&
     {
         if (m_exception_ptr)
         {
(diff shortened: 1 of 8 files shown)


inc/coro/task.hpp
```diff
@@ -173,9 +175,9 @@ class [[nodiscard]] task
     /**
      * @return True if the task is in its final suspend or if the task has been destroyed.
      */
-    auto is_ready() const noexcept -> bool { return m_coroutine == nullptr || m_coroutine.done(); }
+    auto is_ready() const noexcept->bool { return m_coroutine == nullptr || m_coroutine.done(); }
 
-    auto resume() -> bool
+    auto resume()->bool
```
jbaldwin, line 180: Huh, not sure why clang-format did this :(

PR #121, opened 2022-10-06
Upgrade to make the project work with nix

jstranik:
Upgrade the expect dependency.
Add install targets.

The changes were necessary to make the project compiled with the nix build system (latest compilers and libraries).


jbaldwin:
Good change on adding expect as a submodule.

.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "vendor/c-ares/c-ares"]
 	path = vendor/c-ares/c-ares
 	url = https://github.com/c-ares/c-ares.git
+[submodule "vendor/tartanllama/expected"]
```
jbaldwin, line 4: Can you update the root README.md to note that this is now pulled by submodule instead of embedded please?

jstranik: Updated README.md

jbaldwin (approved):

PR #131, opened 2023-03-04
task_container should cleanup tasks inline

### LINKED ISSUE #130: task_container should cleanup tasks inline when they are completed
Issues with creating the asynchronous echo server where the application would run out of sockets.  It was possible to work around this by detecting the `accept()` failture and then running the `task_manager.garbage_collect()` manually but ideally tasks should clean themselves up properly without the user needing to be involved.

The proposed solution is in the `make_cleanup_task()` to invoke the task slot that is being recycled to have its destructor invoked immediately and not when the slot is re-used.  This should also remove the need for the gc functions.

jbaldwin:
* This was an issue identified with the asynchronous echo server where running out of file descriptors would cause tasks to never be cleaned up in the task container.  Cleaning them up inline when they complete should resolve this issue.

Closes #130


jbaldwin:
The issue around this seems to be branch shaking with a static lib, when statically linking some of the coroutine frame.Destroy functions are omitted, so I'll switch to dynamically linking for the forseeable future to fix this issue.

examples/CMakeLists.txt
```diff
@@ -2,73 +2,80 @@ cmake_minimum_required(VERSION 3.0)
...
-    target_compile_options(coro_shared_mutex       PUBLIC -fcoroutines -Wall -Wextra -pipe)
-    target_compile_options(coro_tcp_echo_server    PUBLIC -fcoroutines -Wall -Wextra -pipe)
-    target_compile_options(coro_http_200_ok_server PUBLIC -fcoroutines -Wall -Wextra -pipe)
-elseif(${CMAKE_CXX_COMPILER_ID} MATCHES "Clang")
-    message(FATAL_ERROR "Clang is currently not supported.")
-else()
-    message(FATAL_ERROR "Unsupported compiler.")
-endif()
\ No newline at end of file
+target_compile_features(coro_http_200_ok_server PRIVATE cxx_std_20)
```
jbaldwin, line 53: delete

jbaldwin: done

PR #133, opened 2023-04-06
Remove "lib" prefix for CMake target

chausner:
CMake prepends library targets with "lib", resulting in a file named "liblibcoro.a". The suggested change sets the prefix to an empty string so that the file is named "libcoro.a".


jbaldwin:
lol oops, at least its an easy fix :smile:

jbaldwin (approved):

PR #134, opened 2023-04-06
Allow to resolve dependencies via find_package

chausner:
This adds a new CMake option `LIBCORO_EXTERNAL_DEPENDENCIES` (off by default) which uses `find_package` to resolve dependencies instead of building them from submodules. This is helpful when one needs to have control over how dependencies are built, for example when using package managers like vcpkg or Conan.


jbaldwin:
Great change, thanks for the PR!

jbaldwin (approved):

PR #135, opened 2023-04-11
Mention Conan in README.md

chausner:
This adds a note to README.md about the availability of libcoro via the Conan package manager.

### DIFF
--- a/.githooks/readme-template.md
+++ b/.githooks/readme-template.md
@@ -453,6 +453,9 @@ target_link_libraries(${PROJECT_NAME} PUBLIC libcoro)
 
 ```
 
+##### Conan
+libcoro is available via the [Conan](https://conan.io/center/libcoro) package manager.
+
 #### Tests
 The tests will automatically be run by github actions on creating a pull request.  They can also be ran locally:
 
--- a/README.md
+++ b/README.md
@@ -1123,7 +1123,6 @@ target_link_libraries(${PROJECT_NAME} PUBLIC libcoro)
 ##### FetchContent
 CMake can include the project directly by downloading the source, compiling and linking to your project via FetchContent, below is an example on how you might do this within your project.
 
-
 ```cmake
 cmake_minimum_required(VERSION 3.11)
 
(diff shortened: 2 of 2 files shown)


jbaldwin:
Does this require builds to be made an uploaded to the Conan package manager?

chausner:
The Conan package manager resolves libraries via so-called recipes from [Conan Center](https://conan.io/center/) which is the official default repository. Recipes basically tell Conan where the library source is located and how it can be built. Conan Center also hosts binary builds for packages for a number of platform/compiler combinations. So when a client installs the libcoro package via Conan, it will either download a prebuilt binary from Conan Center if available for the client platform/compiler, or it will build the library on-the-fly using the build instructions in the recipe.

Conan Center is maintained by the community as a repository on GitHub: https://github.com/conan-io/conan-center-index/. Everyone can submit new recipes or recipe updates via Pull Requests. I already did the necessary work to write a recipe for the current version of libcoro: https://github.com/conan-io/conan-center-index/pull/16877.

When a new version of libcoro is published, someone will need to manually update the recipe so that the new version is available on Conan Center. If nothing changed in the build process, it is trivial to update the recipe (basically one only needs to add the GitHub URL of the new version tag). If there are changes in how the library is built, more changes might be needed.
In general, it is not expected that you update it (although you are free to submit recipe updates yourself, if you want). Ideally someone from the Conan community will update the recipe soon after a new release is made but in general, there is of course no guarantee that Conan Center always hosts the latest version of all libraries since it is a community effort.

chausner:
By the way, I also added a port of libcoro to the vcpkg package manager: https://github.com/microsoft/vcpkg/pull/30623. The process there to get library ports updated is very similar. Unfortunately, their CI currently runs gcc 9 which is too old for libcoro, so they decided to not merge my PR for now until they update their CI pipeline. But I hope that sometime in the future, libcoro will also be available on vcpkg.

jbaldwin:
Thanks, I appreciate the in depth explanation.  Seems straight forward enough.  That's too bad on vcpkg!  Hopefully they can get on a more recent compiler 😃

jbaldwin (changes requested):
This change actually needs to be in the readme-template.md file otherwise it will be blown away by the commit process.  It injects all the built examples via macros/tokens so they don't have to be manually copied and pasted.  Can you move it there please?

chausner:
Sure, I copied the changes to `readme-template.md`.

jbaldwin (approved):

jbaldwin:
Thanks!

PR #141, opened 2023-06-02
Add a LIBCORO_FEATURE_NETWORKING option

### LINKED ISSUE #138: OSAL: any help or ideas
TL;DR: Help needed in wrapping the C++ threading, synchronization, POSIX epoll, etc APIs in a new operating system abstraction layer for easy porting to different embedded OSes that have the features, but not the identical APIs.

I am pondering using libcoro on embedded systems, viz. Arduino and Zephyr BSPs. I haven't tried the comprehensive support for POSIX provided by NUTTX yet, given a recent compiler, it might be be possible to build for that OS with only minor tweaks. Anyway, the Zephyr SDK and at least the ESP8266 Arduino Core offer recent enough compiler toolchains to support <coroutine>.

What I am asking if there is any prior art or guidance for how to go about an OSAL in libcoro? Before I start reading all the sources :-) As first step, I have made a branch that (temporarily) purges all networking APIs and tests, but builds the remaining examples for Ubuntu. I should attempt building that for Zephyr native POSIX target in the coming days, if that works, it would be helpful in a step-by-step migration to an OSAL implementation.

jbaldwin:
A request was made for an OSAL type build of libcoro. To accomodate this I opted to turn the networking portion of libcoro into an optional build type that is enabled by default.

Closes #138


dok-net:
Looks good to me. I would find renaming `inc` to `include` a good idea, but that's for another PR, just like further OSAL work concerning mutex, condvar, semaphore, atomic...

jbaldwin:
Cool, I'll probably merge this then since I like the idea of certain aspects being enabled/disabled based on usage, and when other things need to be split apart they can be done in their own tasks.

PR #149, opened 2023-06-05
Allow to disable OpenSSL separately if networking is included.

dok-net:
I found that it's necessary for some targets to omit the SSL support. This is a plain patch with #ifdefs throughout the code.


dok-net:
I think the TCP echo and HTTP server examples could use some explaining in regard to their worker model.

jbaldwin:
This is a great change, I really like the idea of splitting out SSL from networking. I'm probably not going to be around much this week (vacation) so likely will take a look at it next week.

dok-net:
The latest commit, in my drive to support embedded targets with partial C++ standard library support, add the option to en/dis-able all featuring requiring std::mutex and friends.
@jbaldwin I see that `coro::when_all()` works without std::mutex. Why is there no `when_any()`? I haven't looked the code yet, obviously.

jbaldwin:
I've taken a cursory look at this and it looks about right on my phone but I'd like to do a deeper dive (hopefully tomorrow) with a proper screen 😄

Hadn't gotten around to a coro::when_any, but it would be a great addition! I think it might require a cancellation of the tasks that didn't complete first so the library might need to introduce a cancellation token type.  It would be dependent upon the users tasks checking for it and stopping execution though.

jbaldwin:
I think the only thing missing is updating the readme-template with the new options, can you add that in please?

jbaldwin (approved):

jbaldwin:
Doesn't seem to have updated the readme from the githook... We'll get it on the next commit.

PR #152, opened 2023-06-11
Add manifest etc to use libcoro (a subset thereof) as Arduino library for ESP8266

dok-net:
The `coro::task` and `coro::generator` are working on the ESP8266 Arduino core due to their updated gcc version. ESP32 may be coming soonish, depending on the gcc version pulled in by IDF 5.1 once it gets released.

### Steps:
Clone the ESP8266 Core for Arduino from GitHub into the sketchbook directory according to the instructions provided with the core.
Put into `<sketchbook>/hardware/esp8266com/esp8266/platform.local.txt`:
```
build.stdcpp_level=-std=gnu++20
compiler.cpp.extra_flags=-fcoroutines -std=gnu++20
```

Clone this PR into `<sketchbook>/libraries/libcoro`.

Run this sketch on an ESP8266:
```
#include <coroutine>
#include <string>
#include <sstream>
#include <vector>

#include <Schedule.h>
#include <Ticker.h>
#include <SoftwareSerial.h> // in order to pull in circular_queue.h
#include <circular_queue/circular_queue.h>
#include <libcoro.h>

struct resume_later : std::suspend_always
{
	void await_suspend(std::coroutine_handle<> handle)
	{
		schedule_function([handle]()
			{ handle.resume(); });
	}
};

struct co_delay : std::suspend_always
{
	co_delay(uint32_t millis) {
		this->millis = millis;
	}

	Ticker ticker;
	uint32_t millis;

	void await_suspend(std::coroutine_handle<> handle)
	{
		ticker.once_ms_scheduled(millis, [handle]()
			{ handle.resume(); });
	}
};

coro::task<int> millis_async()
{
	co_return millis();
}

coro::task<std::string> noopAsync(int id)
{
	co_await resume_later();
	std::stringstream hello;
	hello << "Hello async" << " (" << id << ") = " << co_await millis_async() << std::ends;
	auto str = hello.str();
	co_await resume_later();
	co_return str;
}

void main_coro(coro::task<void>& task, int id)
{
	task = [](coro::task<void>& task, int id) -> coro::task<void>
	{
		auto result = co_await noopAsync(id);
		Serial.printf("main_coro(%u) = \"%s\"\n", id, result.c_str());
		// restart main_coro, recycling the task object's space from the previous run
		co_await co_delay(id * 1000);
		schedule_function([&task, id]() { main_coro(task, id); });
	}(task, id);
	task.resume();
}

coro::generator<coro::task<void>> main_tasks(int count)
{
	for (int i = 0; i < count; ++i) co_yield coro::task<void>();
}

circular_queue<coro::task<void>> tasksList(10);

void setup()
{
	Serial.begin(115200);
	while (!Serial) delay(1);

	//main_gen();

	int i = 1;
	auto tasks = main_tasks(tasksList.capacity());
	for (auto it = tasks.begin(); it != tasks.end(); ++it)
	{
		tasksList.pushpeek() = std::move(*it);
		main_coro(tasksList.pushpeek(), i);
		tasksList.push();
		++i;
	}
}

void loop()
{
}
```


dok-net:
@mcspr heads up: the marvelous C++20 awaitable stackless coroutines, this great libcoro, and some plumbing to bring it all together, and we easily have 20+ asynchronous tasks running on the ESP8266. Async Webservers, I/O, in one homogenous modern programming pattern. I find it amazing on such small silicon, and I think it's more than just a showcase. Would you consider adding the contents of `platform.local.txt` to the default build, turning off or fixing the warning introduced by that (`volatile`...) in the line of business?

dok-net:
@jbaldwin I know this is quite intrusive, due to Arduino's build system. Discussing it doesn't cause any damage, right?
I don't know of any way to get the cpp's out of "harms way", the Arduino library spec has no concept of a separate "include" directory, it adds "src" to the include path, period. All I can do is prevent it from adding just all headers in src to the sketch/program when a user choses the lib from the menu.

jbaldwin:
Yeah it's fairly intrusive, especially merging the inc and src dirs... I'm guessing you are running into maybe [this](https://forum.arduino.cc/t/how-to-include-from-subfolder-of-sketch-folder/428039)  problem?

I'm also assuming for Arduino you're not using the cmake build system, is that correct?  If so I wonder if a special script or something could prepare the source into the desired directory format for this platform?

dok-net:
The intent is to publish an Arduino library, hence the addition of a `library.properties` file. There are some restrictions to keep, and few to no configuration options for that, like it's not possible to set compiler flags. As a result, all sources including at least one header must be in the top-level `src` directory. If libcoro can export a working Arduino library tree from some CMake command, that would do the job as well. To publish the lib to the Arduino library manager, eventually there has to be a Github repo that gets registered with the library manager, which could be done manually today, but a maintained procedure is preferable, isn't it?

mcspr:
*(too too distracted lately, sry for late replies)*

> special script or something could prepare the source into the desired directory format

> If libcoro can export a working Arduino library tree from some CMake command, that would do the job as well

Would concur, something building arduino library layout from the existing layout would be less intrusive. This action / custom target / command should

1) make libcoro/ the output library dir, and place library.properties there (since arduino can't detect 'flat' layout with .properties included, and **must** have src/ dir as library root)
2) move src/, inc/coro (and relevant vendor/ if needed) to libcoro/src/, and just ignore files not currently used with our platform like networking and stuff
3) make libcoro/src/libcoro.hpp and supply includes=libcoro.hpp via library.properties, single-include for arduino environment (since it does not seem to work with coro/coro.hpp as-is)
```cpp
#pragma once

#ifndef ARDUINO
#error "Arduino-only!"
#endif

#include "coro/coro.hpp"
```
4) .zip it up / copy to custom dir, which is then published to a separate repo or branch for Arduino library manager

Resulting tree is something like this, arduino-cli picks it up when using `#include <libcoro.hpp>`
```console
~/A/l/libcoro> tree
.
├── library.properties
└── src
    ├── coro
    │   ├── concepts
    │   │   ├── awaitable.hpp
    │   │   ├── buffer.hpp
    │   │   ├── executor.hpp
    │   │   ├── promise.hpp
    │   │   └── range_of.hpp
    │   ├── coro.hpp
    │   ├── detail
    │   │   ├── poll_info.hpp
    │   │   └── void_value.hpp
    │   ├── event.hpp
    │   ├── fd.hpp
    │   ├── generator.hpp
    │   ├── io_scheduler.hpp
    │   ├── latch.hpp
    │   ├── mutex.hpp
    │   ├── net
    │   │   ├── connect.hpp
    │   │   ├── dns_resolver.hpp
    │   │   ├── hostname.hpp
    │   │   ├── ip_address.hpp
    │   │   ├── recv_status.hpp
    │   │   ├── send_status.hpp
    │   │   ├── socket.hpp
    │   │   ├── ssl_context.hpp
    │   │   ├── ssl_handshake_status.hpp
    │   │   ├── tcp_client.hpp
    │   │   ├── tcp_server.hpp
    │   │   └── udp_peer.hpp
    │   ├── poll.hpp
    │   ├── ring_buffer.hpp
    │   ├── semaphore.hpp
    │   ├── shared_mutex.hpp
    │   ├── sync_wait.hpp
    │   ├── task_container.hpp
    │   ├── task.hpp
    │   ├── thread_pool.hpp
    │   └── when_all.hpp
    ├── event.cpp
    ├── io_scheduler.cpp
    ├── libcoro.hpp <-- our custom header
    ├── mutex.cpp
    ├── net
    │   ├── connect.cpp
    │   ├── dns_resolver.cpp
    │   ├── ip_address.cpp
    │   ├── recv_status.cpp
    │   ├── send_status.cpp
    │   ├── socket.cpp
    │   ├── ssl_context.cpp
    │   ├── tcp_client.cpp
    │   ├── tcp_server.cpp
    │   └── udp_peer.cpp
    ├── semaphore.cpp
    ├── sync_wait.cpp
    └── thread_pool.cpp
```

https://arduino.github.io/arduino-cli/latest/library-specification/ and builder are pretty limited, only so much can be done
And if some features have to be turned on or off via build flags, it can be pretty tricky to apply properly since Arduino build environment does not really want you changing them :/

> Would you consider adding the contents of platform.local.txt to the default build

See https://github.com/esp8266/Arduino/issues/8916? Or open up a separate issue related to -std=c++20 issue(s) or its inclusion by default
afaik .local.txt is not supposed to be distributed with the Core, so it either has to be our default or extra work for the user to supply it manually. And either way, off-topic here :>

jbaldwin:
For (4) I think a `main-ardunio` branch is appropriate.  Do you think this can all be automated via GitHub actions?  We could do a post merge into `main` GitHub trigger that rebuilds the Arduino branch.  Only downside is if it fails the main and Arduino branches are broken until a user fixes it.  Otherwise I think your outlined plan looks good.

dok-net:
I am withdrawing this PR due to an own new implementation as part of ghostl, in the circular_queue part of EspSoftwareSerial and friends. I intend to turn that into an own project instead of reusing the sources by copy in currently 3 of my projects. It has outgrown that distribution model by far now.
@jbaldwin Thank you for the invaluable help.

PR #154, opened 2023-06-14
Extended tasks so they can return references and not just values.

bruno-j-nicoletti:
I needed the ability to return a reference to an object from a task, rather than copying the object. So I extended tasks to do that and added two tests for it.

There seems to be some reformatting spam in task.h from clang-format being run.

### DIFF
--- a/inc/coro/task.hpp
+++ b/inc/coro/task.hpp
@@ -60,27 +60,48 @@ protected:
 template<typename return_type>
 struct promise final : public promise_base
 {
-    using task_type        = task<return_type>;
-    using coroutine_handle = std::coroutine_handle<promise<return_type>>;
+    using task_type                                = task<return_type>;
+    using coroutine_handle                         = std::coroutine_handle<promise<return_type>>;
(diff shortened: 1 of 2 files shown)


jbaldwin:
For your use case can you return a value type that wraps the ref? My worry here by allowing users to return ref types is that it becomes extremely easy to return a dangling reference on the coroutine frame which is immediately invalid upon the task completing.  It's certainly still possible to do so today but it's a bit harder.

I see in one of the tests it is returning a ref to an object that is created before the coroutine which is technically a valid case, but I think it could be worked around by introducing a small wrapper class to the red, e.g.

```
struct Wrapper
{
    int& result
}

```

Would this work for your use case?

jbaldwin:
Yeah I'm not opposed to it, just figured it wasn't necessary a long time ago since there are ways around it but there are certainly good reasons to do it and it is a part of the language.  Like anything else with c++ user beware... And using std::reference_wrapper is a typing pain too, it's so long!

jbaldwin (approved):

PR #156, opened 2023-06-21
Fix #154 that broke task.hpp by using optional without adding #include

dok-net:
@jbaldwin Please check that the README get regenerated, it didn't last time, and I have disabled that commit hook because it's not working here with my setup yet.

### DIFF
--- a/inc/coro/task.hpp
+++ b/inc/coro/task.hpp
@@ -4,6 +4,7 @@
 
 #include <coroutine>
 #include <exception>
+#include <optional>
 #include <utility>
 
 namespace coro


jbaldwin:
Ok, probably I will make a standalone commit to update it then since it's out of date at the moment.

I assume this fix is for Arduino since all the CI pipelines passed?

dok-net:
> I assume this fix is for Arduino since all the CI pipelines passed?

This reveals that it might be helpful to run CI also with all options disabled. It didn't occur exactly with Arduino - it would have been noticed there, too, to be sure - but in the CMake based use as a library module for Zephyr OS applications.

The upcoming Espressif IDF 5.1 and the Arduino core for the ESP32 family uses a recent GCC such that libcoro now works there, too, not only on the ESP8266. From what I read, CMake is or will be the default build system for their IDF.

jbaldwin:
Aha! That makes perfect sense. Would you mind opening a bug to add that path to CI?

PR #159, opened 2023-06-30
Fix wrong version number

MichaelLettrich:
The version does not correspond to the git tag.


jbaldwin:
Oh fun, I need to add this to the checklist when bumping a version.  Thanks for updating it.

jbaldwin (approved):

PR #161, opened 2023-07-05
New task_completion_source

dok-net:
I am proposing this for inclusion. It's made from the existing building blocks, viz. std::suspend_always. It is actually independant of coro::task etc. but should be fully interoperable. It is useful for implementing async queues, driver event patterns, anything that needs to resume (and pass data to) async functions waiting for some event. This may overlap with the schedulers that libcoro provides, but these are complete solutions and rely on broader OS support for epoll and mutexes etc, whereas this implementation is lock-free.

Using it goes like this:
```
auto eventAsync()
{
  coro::task_completion_source<int> tcs; // give tcs to some context or scheduler, even non-async
  // ...
  return tcs.token(); // co_await'ing tcs.token() continues or suspends depending on when tcs.set_value(42) is called
}
void someOtherContext()
{
  // ...
  // if co_wait tcs.token() has already suspended before tcs.set_value() gets called, it needs to be
  // resumed from another context or scheduler. use tcs.handle() to obtain the coroutine_handle<> for this.
  if (auto handle = tcs.handle(); handle && !handle.done()) {
    handle.resume();
  }
  // ...
}
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,6 +62,7 @@ set(LIBCORO_SOURCE_FILES
     inc/coro/coro.hpp
     inc/coro/generator.hpp
     inc/coro/task.hpp
+    inc/coro/task_completion_source.hpp
     inc/coro/when_all.hpp
 )
 
--- a/inc/coro/coro.hpp
+++ b/inc/coro/coro.hpp
@@ -38,4 +38,5 @@
 
 #include "coro/generator.hpp"
 #include "coro/task.hpp"
+#include "coro/task_completion_source.hpp"
 #include "coro/when_all.hpp"
--- a/inc/coro/task_completion_source.hpp
+++ b/inc/coro/task_completion_source.hpp
@@ -8,8 +8,22 @@ template<typename T = void>
 struct task_completion_source
 {
 private:
-    coro::task<>     suspend() { co_await std::suspend_always(); }
-    coro::task<>     suspended{suspend()};
+    struct suspend
+    {
+        std::atomic<bool>                    ready{false};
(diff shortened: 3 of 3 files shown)


dok-net:
ATM, this doesn't work if `setValue()` gets called before `task()` becomes co_awaited. There may be a fix using `std::suspend_never`, but I am not sure if this will be completely safe against race-conditions. But perhaps multi-threading isn't important, as long as a synchronous `setValue()` in the execution path of setting up the TCS doesn't fail.

jbaldwin:
Can you add some tests and examples to the readme+examples/?  I'm still not quite understanding the use case here and I think that would be helpful to elucidate why this should be added to the library.

This kind of looks like `coro::event` in some regards? But it seems like you are trying to make a construct that works outside of `coro::thread_pool` or `coro::io_scheduler`?

dok-net:
I think I have nailed the concurrency issue.
I'm baffled that a task_completion_source needs explaining, but that's probably because I have been following async on the .NET platform since day 0.
I will add something to the README. In short, there are two things it provides which I haven't found in libcoro so far.  Injecting data from a different context - isn't event "void"? Controlling resumption from a non-async context, so yes, I find libcoro most useful if I am not bound to the provided schedulers, not just because they don't work on my targets, but because I really need the freedom to build my own. Besides, more building blocks instead of a closed solution looks more promising for long-term use instead of having to migrate too soon.

dok-net:
Updated the introductory comment to match the latest revision.

dok-net:
I am closing this PR without further intent on merging, due to completion of a separate async coroutine library of my own making. Generally, if done right, all libs should be interoperable, no?

PR #174, opened 2023-09-29
Use std::expected when __cpp_lib_expected is defined

### LINKED ISSUE #145: Use regular standard lib std::expected if available (C++23)
Early C++23 support be some compilers includes std::expected already.

jbaldwin:
This doesn't appear to actually be working yet due to g++-12 and g++-13 not setting any values higher than __cplusplus 202002L even when passing in -std=c++23.  The code works when manually enabled though. Hopefully a new version of g++ will set this properly and it will just work.

Closes #145

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -34,7 +34,7 @@ if(LIBCORO_FEATURE_THREADING)
 endif()
 
 add_executable(${PROJECT_NAME} main.cpp ${LIBCORO_TEST_SOURCE_FILES})
-# target_compile_features(${PROJECT_NAME} PUBLIC cxx_std_20)
+target_compile_features(${PROJECT_NAME} PUBLIC cxx_std_20)
 target_include_directories(${PROJECT_NAME} PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
 target_link_libraries(${PROJECT_NAME} PUBLIC libcoro)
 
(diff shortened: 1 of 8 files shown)


test/CMakeLists.txt
```diff
@@ -29,7 +29,7 @@ if(LIBCORO_FEATURE_NETWORKING)
 endif()
 
 add_executable(${PROJECT_NAME} main.cpp ${LIBCORO_TEST_SOURCE_FILES})
-target_compile_features(${PROJECT_NAME} PUBLIC cxx_std_20)
+# target_compile_features(${PROJECT_NAME} PUBLIC cxx_std_20)
```
jbaldwin, line 32: delete me

jbaldwin: un commented

bjones1:
Could you test for this a [feature test](https://en.cppreference.com/w/cpp/feature_test) instead? Perhaps

``` C++
#ifdef __cpp_lib_expected
   #include <expected>
#else
    #include <coro/expected.hpp>
#endif
```

jbaldwin:
Great idea, didn't know these existed. I'll give it a shot!

jbaldwin:
I've switched it to use `__cpp_lib_expected`, its probably good enough to move forward with.

jbaldwin:
This change also has some readme updates based on features recently merged.

Issue #170, opened 2023-09-20
Proposed changes to threading option


bjones1:
Since I'm working on getting libcoro ported to other compilers/platforms, would you be open to a redefinition of the `LIBCORO_FEATURE_THREADING` option in `CMakeLists.txt`? Current, the docstring there is "Include multithreading features." Could we make this "Include *platform-specific* multithreading features", and include all features that are platform-independent for all platforms? Almost all of the current threading features will work on Windows (and probably all platforms).

jbaldwin:
Would this require a `LIBCORO_FEATURE_THREADING_WINDOWS|LINUX` cmake option if I'm understanding correctly?

What about detecting the compiler/platform and including the appropriate features that way and keeping the option named the same?  That way the user doesn't have to care about which platform they are on for the option, if its enabled it'll include as much as possible that that platform supports.  Might be good to document _what_ features each platform would include in the readme

bjones1:
> Would this require a `LIBCORO_FEATURE_THREADING_WINDOWS|LINUX` cmake option if I'm understanding correctly?

Instead, I was thinking of making the `LIBCORO_FEATURE_THREADING` include Linux-only features, and including everything that's platform-agnostic by default.

> What about detecting the compiler/platform and including the appropriate features that way and keeping the option named the same? That way the user doesn't have to care about which platform they are on for the option, if its enabled it'll include as much as possible that that platform supports. Might be good to document what features each platform would include in the readme

I agree that we should automatically exclude features that aren't supported on the OS used to build the library. But I also think that most of the libcoro features should be included by default; after all, as a library, developers can simply link against what they want; there's little penalty for including more. In addition, a large amount of `libcoro` is implemented as header files, so there's no executable code to include in a library.

Here's a concrete proposal for making most of the `LIBCORO_FEATURE_THREADING` an unconditional part of the build, leaving only Linux-specific files behind that feature. This could be improved by future PRs supporting other OSes.

Files that are currently in `LIBCORO_FEATURE_THREADING` that would be an unconditional part of the build:
- include/coro/concepts/awaitable.hpp
- include/coro/concepts/buffer.hpp
- include/coro/concepts/executor.hpp
- include/coro/concepts/promise.hpp
- include/coro/concepts/range_of.hpp

-  include/coro/detail/void_value.hpp

-  include/coro/coro.hpp
- include/coro/event.hpp src/event.cpp
- include/coro/generator.hpp
- include/coro/latch.hpp
[...]

PR #175, opened 2023-09-30
Fix: Move platform-agnostic files outside of `LIBCORO_FEATURE_THREADING`

bjones1:
This implements the suggestion in #170.


CMakeLists.txt
```diff
@@ -113,7 +112,7 @@ endif()
 add_library(${PROJECT_NAME} STATIC ${LIBCORO_SOURCE_FILES})
 set_target_properties(${PROJECT_NAME} PROPERTIES LINKER_LANGUAGE CXX PREFIX "")
 target_compile_features(${PROJECT_NAME} PUBLIC cxx_std_20)
-target_include_directories(${PROJECT_NAME} PUBLIC include)
+target_include_directories(${PROJECT_NAME} PUBLIC include vendor/tartanllama/expected/include)
```
jbaldwin, line 115: I'm curious why this was needed? Is this an MSVC problem? It doesn't feel like the right approach, especially since I'm working on a branch to use `std::expected` if the user has a C++23 compiler with support.

test/CMakeLists.txt
```diff
@@ -28,6 +22,17 @@ if(LIBCORO_FEATURE_NETWORKING)
     )
 endif()
 
+if(LIBCORO_FEATURE_THREADING)
```
jbaldwin, line 25: What would you consider a good name for this option variable moving forward? `THREADING` doesn't quite seem to make sense to me anymore but I'm not sure what it _should_ be.  Do you have any suggestions?

jbaldwin (changes requested):

CMakeLists.txt
bjones1: Several files (`semaphore.hpp`, `ring_buffer.hpp`) have the statement `#include <tl/expected.hpp>`. I don't understand how this builds without this change.

Another option is to use the C++23 std::expected, but I still get errors on MSVC when I change this to `#include <expected>` and change `tl::expected` to plain `expected` in `ring_buffer.hpp`:

```
ring_buffer.hpp(149): error C7568: argument list missing after assumed function template 'expected'
ring_buffer.hpp(161): note: see reference to class template instantiation 'coro::ring_buffer<element,num_elements>::consume_operation' being compiled
```

bjones1: However, I'd be quite happy to use the C++23 approach in your branch if that works.

bjones1: I'd be happy to revert this line if you'd like; I think the rest of the PR is OK. We can come back to this.

test/CMakeLists.txt
bjones1: Let me think a bit on this...

bjones1: I still like `THREADING` -- at its heart, this is the `io_scheduler.hpp/cpp` that schedules coroutines across multiple threads; everything else (from what I see) supports this. The items moved outside of `THREADING` don't belong in threading, since they support coroutines, not threads.

CMakeLists.txt
jbaldwin: I think it is unnecessary.  If `LIBCORO_EXTERNAL_DEPENDENCIES=OFF` then the root CMakeLists.txt  around line 37 to 54 includes `add_subdirectory(vendor/tartanllama/expected)` which will set the include directory appropriately for `#include <tl/expected>`.  That feature was added a while ago so users could provide their own if they wanted to avoid git submodules or already had them in their project so they wouldn't have to be doubly included.  The user needs to set include paths correctly in _their_ build system though if they do this.

-----

The PR I'm working on for `std::expected` is here: https://github.com/jbaldwin/libcoro/pull/174 you can see how I made a compile time interface to pick one or the other... but the compilers don't seem to be setting `__cplusplus` versions properly so the build is failing :(

test/CMakeLists.txt
jbaldwin: Counterpoint: You can run `io_scheduler` on 1 thread and inline coroutine execution on the `io_scheduler`'s event loop thread instead of having a `coro::thread_pool` to execute ready coroutines.  I also think the _threading_ part isn't the issue with cross platform, its the kernel events (e.g. epoll) that are the real problem right now since its basically hardcoded to linux epoll where as `std::thread` should be portable.  Something like... `LIBCORO_FEATURE_IO` might make sense?

CMakeLists.txt
bjones1: > I think it is unnecessary. If `LIBCORO_EXTERNAL_DEPENDENCIES=OFF` then the root CMakeLists.txt around line 37 to 54 includes `add_subdirectory(vendor/tartanllama/expected)` which will set the include directory appropriately for `#include <tl/expected>`. `LIBCORO_EXTERNAL_DEPENDENCIES` feature was added a while ago so users could provide their own if they wanted to avoid git submodules or already had them in their project so they wouldn't have to be doubly included. The user needs to set include paths correctly in _their_ build system though if they do this.

OK, thanks -- I missed that when playing with the library. I'll update my PR to remove this change.

test/CMakeLists.txt
bjones1: That's true, it's the epoll dependency that causes problems.

bjones1:
Do you want me to merge in the `main` branch, or rebase this branch to `main`? I'm happy to, just let me know your workflow.

Is there anything else to do on this PR?

test/CMakeLists.txt
jbaldwin: https://github.com/jbaldwin/libcoro/issues/183

jbaldwin (approved):

jbaldwin:
Generally whatever works best, I usually only squash and rebase into main so it's a singular commit in the final git log.

bjones1:
Looks like you merged main in successfully -- thanks! Do you need anything else from me on this PR?

I'm excited to get this merged; I *think* this will get us to successful builds on XCode and MSVC. I'll do some more testing/cleanup after the merge.

test/CMakeLists.txt
```diff
@@ -28,6 +22,17 @@ if(LIBCORO_FEATURE_NETWORKING)
...
+    list(APPEND LIBCORO_TEST_SOURCE_FILES
+        bench.cpp
+        test_semaphore.cpp
```
jbaldwin, line 28: I see there is an old usage of `coro::io_scheduler` in here for using `yield()` but I added `yield()` onto `coro::thread_pool` recently, how would you feel about updating this test so this file could be moved outside of the `LIBCORO_FEATURE_THREADING` flag?

test/CMakeLists.txt
```diff
@@ -28,6 +22,17 @@ if(LIBCORO_FEATURE_NETWORKING)
...
+        test_io_scheduler.cpp
+        test_ring_buffer.cpp
```
jbaldwin, line 30: I believe the usage of `coro::io_scheduler` in this file can also be replaced with `coro::thread_pool` 1 thread

test/CMakeLists.txt
```diff
@@ -28,6 +22,17 @@ if(LIBCORO_FEATURE_NETWORKING)
...
+        test_shared_mutex.cpp
+        test_thread_pool.cpp
```
jbaldwin, line 32: This file seems like it can be moved out of the feature? I don't see any usages of `coro::io_scheduler` in it.

test/CMakeLists.txt
```diff
@@ -28,6 +22,17 @@ if(LIBCORO_FEATURE_NETWORKING)
     )
 endif()
 
+if(LIBCORO_FEATURE_THREADING)
+    list(APPEND LIBCORO_TEST_SOURCE_FILES
+        bench.cpp
+        test_semaphore.cpp
+        test_io_scheduler.cpp
+        test_ring_buffer.cpp
+        test_shared_mutex.cpp
```
jbaldwin, line 31: This file probably cannot be moved easily, but perhaps the best we could do is maybe make the test that uses `coro::io_scheduler` be wrapped in a `#define` for the feature so not all of the tests are disabled?

jbaldwin (changes requested):

jbaldwin:
I was planning on merging this tonight, but I did take one last look at it and I do think _some_ of the test files can probably be moved outside of the feature flag pretty easily by switching the tests from `coro::io_scheduler` to `coro::thread_pool`.  Unless if there is another reason you included them inside the feature flag?  Would you mind taking a look and letting me know what you think.

test/CMakeLists.txt
bjones1: Nice! I'm always in favor of including more tests. Done.

test/CMakeLists.txt
bjones1: I wasn't able to get this one working -- the program hung:

```
22:21:31: Starting test\libcoro_test.exe...
Randomness seeded to: 2625223567
Acquiring lock
lock acquired, emplacing back 1
coroutine done
block task acquiring lock
block task acquired lock, waiting on event
id = 1 waiting to acquire the lock
id = 2 waiting to acquire the lock
id = 3 waiting to acquire the lock
id = 4 waiting to acquire the lock
set task setting event
id = 4 lock acquired
id = 4 coroutine done
id = 3 lock acquired
id = 3 coroutine done
id = 2 lock acquired
id = 2 coroutine done
id = 1 lock acquired
id = 1 coroutine done
produce: 1
produce: 2
produce: 3
consume: 1
produce: 4
consume: 2
produce: 5
consume: 3
produce: 6
consume: 4
produce: 7
consume: 5
produce: 8
consume: 6
produce: 9
consume: 7
produce: 10
consume: 8
consume: 9
consume: 10
```

test/CMakeLists.txt
bjones1: Thanks, done!

test/CMakeLists.txt
bjones1: Fantastic, done!

test/CMakeLists.txt
jbaldwin: Interesting, I'll mark it as something to look at deeper, we can keep it as is for now.

bjones1: Oops, I think I was just impatient. It does pass, but takes a while to run. Does this sound OK to you?

jbaldwin: It does do 10 million produce and consumes, so should take a little time but not a crazy amount.. if you run it a few more times is it consistently long? It's possible the yield is choking it out and the 50ms yield previously was making it complete a lot faster...

bjones1: Here's an example from CI (the Ubuntu 22.04 run):

```
Thu, 12 Oct 2023 03:32:59 GMT 2: consume: 10
Thu, 12 Oct 2023 03:34:04 GMT 2: ===============================================================================
```

It takes about 1 minute to run this last line on Ubuntu.

jbaldwin: I'd maybe run it before and after the switch and time it, that'll tell us if we made a major regression in the speed.

bjones1: Before the switch on my PC, it takes a few seconds at most. After, a while (guessing 1 minute?).

bjones1: I can put this in a separate PR if you'd like to look at it. For now, it's in a separate branch: https://github.com/bjones1/libcoro/tree/slow-threading-refactor.

bjones1: Here's the CI from Fedora-37; again, about a minute:

```
Thu, 12 Oct 2023 03:32:36 GMT 2: consume: 10
Thu, 12 Oct 2023 03:33:24 GMT 2: ===============================================================================
```

bjones1: Looking at [earlier CI where this test was still included](https://github.com/jbaldwin/libcoro/actions/runs/6461866284/job/17542416231), it looks like that's a pretty standard time:

```
Mon, 09 Oct 2023 21:30:13 GMT 2: consume: 10
Mon, 09 Oct 2023 21:31:05 GMT 2: Acquiring semaphore
```

bjones1: So this should be fine to include. I'll go ahead and add it to the PR.

jbaldwin: Ok great, thanks for checking

jbaldwin (approved):

Issue #166, opened 2023-09-19
Support for other compilers?


bjones1:
Thanks for the great library! It's enabled me to do some interesting work with coroutines without the huge amount of work (which you've done) needed to make them useful in C++20.

I'm using parts of libcoro on other compilers -- XCode and MSVC. Would you be interested in small PRs to improve support for these compilers? I recognize that large parts (networking, etc.) are very tied to Linux, and don't plan on fixing that. However, a few tweaks (missing include files MSVC reports, a few syntax updates from XCode, etc.) would certainly benefit me.

Thanks again!

jbaldwin:
Hey, glad you're are finding libcoro useful.  I think expanding the compiler and platform support at this point is an excellent goal. You're right a lot of the networking code is 100% Linux oriented so we'd probably just need to disable compiling that for now for non Linux platforms.

Small support for more common coroutine items would be very welcome, I'm working on another pr now to get clang 16 supported with another user so this feels like a natural next step as well.

bjones1:
Fantastic! I've submitted #169, and will follow with other PRs.

A question: there are a few places in the code where you use an `__attribute__((used))`. For other compilers, would you like me to make this a macro that's empty for compilers that don't understand it, or would you rather just drop this?

Thanks!

bjones1:
I'm leaning toward keeping `__attribute__((used))`. If so, what do you think about a new header file named `predef.hpp`, which would be something like:

``` C++
#pragma once
#ifdef __GNUC__
#define __ATTRIBUTE__USED__ __attribute__((used))
#else
#define __ATTRIBUTE__USED__
#endif
```

Does that work for you?

jbaldwin:
Apologies, I somehow missed your latest comment on here, looks like you opened a PR and we worked through it anyways :smile: 

Good if we close this issue out?

bjones1:
Absolutely!

PR #178, opened 2023-10-07
Fix: Define macros for __attribute__.

bjones1:
Since `__attribute__` is GCC-only, provide a way to omit on other compilers. See also the discussion in #166.


bjones1:
Thanks!

PR #179, opened 2023-10-08
Minor fix for compatibility with OpenSSL < 1.1.0

Harvey-Liu:
1、Minor fix for compatibility with OpenSSL < 1.1.0
2、Examples make sure the socket can be read/write before recv/send


jbaldwin:
Thanks for the PR! I think I left off the polls originally to make the examples simple but you're correct, any real program should be checking if they are readable or writable.

jbaldwin (approved):

PR #185, opened 2023-10-12
Fix: Get examples building on compilers that don't support threading.

bjones1:


jbaldwin:
Looks like this one also needs a merge conflict resolution.

jbaldwin (approved):

bjones1:
You bet, done!

PR #200, opened 2023-10-27
task promise use std::variant

### LINKED ISSUE #196: Explore std::variant for coro::task return value / exception
@a858438680 committed a nice change to use less memory in coro::task, I'd like to explore using a more type safe std::variant since we are on C++20 instead of `unsigned char[]` directly.  I'd want to test to make sure the size difference is the same, but I think it'll reduce some of the metaprogramming to determine which type is larger and allocating based off of that.

jbaldwin:
This is a step towards using a more modern container within the task promise and reduce some manual overhead around managing the unsigned char[] that would store the return value but retaining the benefits that that solution brought to the task's promise class.

Closes #196

### DIFF
--- a/include/coro/task.hpp
+++ b/include/coro/task.hpp
@@ -4,6 +4,7 @@
 #include <exception>
 #include <stdexcept>
 #include <utility>
+#include <variant>
 
 namespace coro
 {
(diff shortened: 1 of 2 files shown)


jbaldwin:
@a858438680 if you have some time I'd appreciate a review on this more modern take for the task optimizations.

PR #201, opened 2023-11-02
Changes to be able to build to webassembly with emscripten.

bruno-j-nicoletti:
The main changes to get it to build with emscripten were...
- copy the single tl::expected header into the repository to avoid transient build issues with catch2,
- use std::threads as opposed to std::jthreads when building with em++ as it does not yet support jthreads,
- tweaked CMakelists to disable  LIBCORO_FEATURE_PLATFORM if building with emscripten and to set appropriate em++ flags.

To build it you need to download the emsdk and set it up you shell's paths appropriately. Then go...
```
mkdir build-emscripten-release
cd build-emscripten-release
emcmake cmake -GNinja -DCMAKE_BUILD_TYPE=Release ..
ninja
```


include/coro/thread_pool.hpp
```diff
@@ -215,7 +218,11 @@ class thread_pool
     /// The configuration options.
     options m_opts;
     /// The background executor threads.
+#ifdef __cpp_lib_jthread
     std::vector<std::jthread> m_threads;
+#else
```
jbaldwin, line 223: I'm kind of thinking for easy of maintenance we just stick with std::thread and manually handle the stop, std::jthread is nice that it encapsulates it but I think the code will be cleaner and easier to maintain if we stick to one solution.  Unless if you think in the future emscripten will support std::jthread and we can bump the required version of that and just support std::jthread from that point moving forward? What do you think?

include/coro/expected.hpp
```diff
@@ -11,7 +11,7 @@ template<typename E>
 using unexpected = std::unexpected<E>;
 } // namespace coro
 #else
-    #include <tl/expected.hpp>
+    #include "coro/detail/tl_expected.hpp"
```
jbaldwin, line 14: ok cool, this is a good solution, and works great moving forward with the std::expected change.

Can you update the readme-template to exclude this as a submodule dependency please?

CMakeLists.txt
```diff
@@ -111,16 +110,24 @@ if(LIBCORO_FEATURE_NETWORKING)
...
 set_target_properties(${PROJECT_NAME} PROPERTIES LINKER_LANGUAGE CXX PREFIX "")
 target_compile_features(${PROJECT_NAME} PUBLIC cxx_std_20)
 target_include_directories(${PROJECT_NAME} PUBLIC include)
-if(MSVC)
-    # Not sure why this is only needed on Windows.
-    target_include_directories(${PROJECT_NAME} PUBLIC vendor/tartanllama/expected/include)
-endif()
+
 if(LIBCORO_FEATURE_PLATFORM)
-    target_link_libraries(${PROJECT_NAME} PUBLIC pthread tl::expected)
```
jbaldwin, line 123: did pthread get added somewhere else consistently? I only see it for EMSCRIPTEN

CMakeLists.txt
```diff
@@ -111,16 +110,24 @@ if(LIBCORO_FEATURE_NETWORKING)
     endif()
 endif()
 
+if(DEFINED EMSCRIPTEN)
+  add_compile_options(-fwasm-exceptions)
```
jbaldwin, line 114: can these be `target_` for the library? Usually isn't bet cmake practice to set things globally

jbaldwin (changes requested):

include/coro/expected.hpp
bruno-j-nicoletti: Will do

CMakeLists.txt
bruno-j-nicoletti: Oops. I'll fix that.

CMakeLists.txt
bruno-j-nicoletti: Will do

include/coro/thread_pool.hpp
bruno-j-nicoletti: I'm assuming that the emscripten folk would support std::jthreads at some point. I'm really surprised it's not working already as I would have thought its only a matter of recompiling the C++ standard library. It's up to you how you want to work it until they do support std::jthreads. I'm fine either way.

CMakeLists.txt
bruno-j-nicoletti: Actually, they need to be enabled globally for all things compiled with emscripten otherwise it will fail. They really are globals.

include/coro/thread_pool.hpp
jbaldwin: Sorry for taking so long to get back to this, been insanely busy.

I think we'll keep it as is and once emscripten supports jthread we'll remove the std::thread solution.

jbaldwin (approved):

jbaldwin:
I've ignored the file in codacy:
`include/coro/detail/tl_expected.hpp`

jbaldwin:
```bash
git clone https://github.com/emscripten-core/emsdk.git
cd emsdk
./emsdk install latest
./emsdk activate latest
source ./emsdk_env.sh
```

build per bruno's steps and then run tests via:
`node ./test/libcoro_test.js`

jbaldwin:
@bruno-j-nicoletti can you add one more item to this PR which is adding a CI run to build and compile with emscripten and run the tests please? I was able to get it to work locally so it shouldn't be too difficult to automate it.

bruno-j-nicoletti:
I've added a CI pass for webassmbly and I've had to tweak test_task.cpp as em++ seems to be adding padding to std::detail::promise

README.md
```diff
@@ -61,7 +61,7 @@ int main()
 
     // Create a task that awaits the doubling of its given value and
     // then returns the result after adding 5.
-    auto double_and_add_5_task = [&](uint64_t input) -> coro::task<uint64_t> {
+    auto double_and_add_5_task = [${EXAMPLE_CORO_TASK_CPP}](uint64_t input) -> coro::task<uint64_t> {
```
jbaldwin, line 64: looks like something weird happened to the README.md file? does it need to be regenerated?

jbaldwin (changes requested):

README.md
bruno-j-nicoletti: That's weird thing I noticed. It keep happening on a commit due to a pre-commit hook. I'll fix it when I get back on my linux box tomorrow.

bruno-j-nicoletti: Done

jbaldwin (approved):

jbaldwin:
I'll rebase once it completes and run the test suite one more time.  This is an awesome change :)

jbaldwin:
There seems to be a new flaky test thats been introduced, doesn't really seem to be anything related to your change osit.

bruno-j-nicoletti:
The only change I could have made that might have done this is the flip to use std::thread over std::jthread. I could put the compile time switch back in to see if that works.

jbaldwin:
Yeah, I really don't see why your change is causing this. I'm going to try and replicate locally...

jbaldwin:
running ubuntu 22.04 with g++-12 dosen't hang like it is here, I might try in a docker container with more constrained resourcecs... but its weird too because other PRs based off `main` are passing just fine

src/thread_pool.cpp
```diff
@@ -75,20 +72,20 @@ auto thread_pool::shutdown() noexcept -> void
...
-auto thread_pool::executor(std::stop_token stop_token, std::size_t idx) -> void
+auto thread_pool::executor(std::size_t idx) -> void
 {
     if (m_opts.on_thread_start_functor != nullptr)
     {
         m_opts.on_thread_start_functor(idx);
     }
 
-    while (!stop_token.stop_requested())
+    while (!m_shutdown_requested.load(std::memory_order::relaxed))
```
jbaldwin, line 82:
```suggestion
    while (!m_shutdown_requested.load(std::memory_order::acquire))
```

src/thread_pool.cpp
```diff
@@ -75,20 +72,20 @@ auto thread_pool::shutdown() noexcept -> void
...
     {
         // Wait until the queue has operations to execute or shutdown has been requested.
         while (true)
         {
             std::unique_lock<std::mutex> lk{m_wait_mutex};
-            m_wait_cv.wait(lk, stop_token, [this] { return !m_queue.empty(); });
+            m_wait_cv.wait(lk, [this] { return !m_queue.empty() or m_shutdown_requested; });
```
jbaldwin, line 88:
```suggestion
            m_wait_cv.wait(lk, [this] { return !m_queue.empty() || m_shutdown_requested.load(std::memory_order::acquire); });
```

src/thread_pool.cpp
jbaldwin: In theory the unique lock should make the `acquire` here unnecessary but its possible the `!m_queue.empty()` triggers us out of the condition variable and hen the `m_shutdown_requested` is never refreshed so the `relaxed` could read stale data.  Lets update these two lines and see if it still reproduces please.

jbaldwin (changes requested):

src/thread_pool.cpp
jbaldwin: My best guess at this point is in the test its not exiting the outer loop properly and hangs.

jbaldwin:
`TEST_CASE("ring_buffer many elements many producers many consumers", "[ring_buffer]")`
This is most likely the test case failing, its using a `ring_buffer` with a `thread_pool` and has a high level of concurrency.

src/thread_pool.cpp
bruno-j-nicoletti: I've made the changes and pushed it, and now it randomly hangs on the CI server for clang 17, g++ and emsdk. They all work fine on my local machine. No idea either.

jbaldwin:
`main` ci has passed twice no problems, going to run it a 3rd time.

fedora 36 has a dnf package manager problem atm, so until they fix that it will fail everytime.

I've looked over the new failing ones and they've failed on a new spot... I'll try and take a deeper look soon.

jbaldwin:
3rd time also passed, just weird package manager problems :roll_eyes:

jbaldwin:
So I had a bunch of fun with this today and I think I found the bug. It was what I originally thought but `acquire` and `release` don't quite work how I thought they did :frowning:. If you switch to `seq_cst` I believe your PR _should_ start passing, but there are two other places that need it as well, in the `io_scheduler.shutdown()` when it exchanges the `m_shutdown_requested` and the same place in the `thread_pool::shutdown()` with the same variable name.  So those two and the the two I already had you change in the `executor` function.  I can push a mirrored PR of your branch with the changes if that is easier.

bruno-j-nicoletti:
I've made the changes you suggested (could you double check please) as well as rebasing. I made it spit out XML so we can see what tests are hanging. This caused a SEGV on top of hanging. It still hangs in the following tests...

- 3 on "semaphore ringbuffer many producers and consumers"
- 1 on "ring_buffer many elements many producers many consumers"
- 2 in "when_all use std::ranges::view".

On Fedora there is a SEGV during coverage tests in '"benchmark tcp_server echo server thread pool".

I cant see a pattern in any of that.

jbaldwin:
I stepped through it in gdb and the `m_shutdown_requested` field was getting set from the test, but wasn't propagating to the other thread_pool threads for the tests that it is hanging on, so the thread_pools never shutdown.  Thats why using seq_cst should in theory fix it.. but clearly theres something deeper wrong in the solution that was being hidden.

The changes you made look right for what I tested locally. I was able to sometimes reproduce if I ran it enough times and seq_cst seemed to fix it, but clearly the github runners seem to disagree.

I'm going to open another PR that only does the seq_cst and std::thread change so we can maybe isolate it a bit more.

jbaldwin:
I've got a working version of `coro::thread_pool` using `std::thread` [here](https://github.com/jbaldwin/libcoro/pull/221).  We can merge that in and then you can rebase, I'm pretty confident this is the issue and should resolve the problems on your PR.

jbaldwin:
Great, thanks for looking.  I've merged it if you want to rebase and :crossed_fingers: this PR starts working!

src/io_scheduler.cpp
```diff
@@ -183,7 +183,7 @@ auto io_scheduler::poll(fd_t fd, coro::poll_op op, std::chrono::milliseconds tim
 auto io_scheduler::shutdown() noexcept -> void
 {
     // Only allow shutdown to occur once.
-    if (m_shutdown_requested.exchange(true, std::memory_order::acq_rel) == false)
+    if (m_shutdown_requested.exchange(true, std::memory_order::seq_cst) == false)
```
jbaldwin, line 186: I think we can back these out since the `notify_all()` was the problem.  Can you revert all of these seq_cst changes back to what they were please?

README.md
jbaldwin, line 64: I'm not sure why your pre-commit is doing this :grimacing: but it looks like it did it again :(

src/thread_pool.cpp
```diff
@@ -51,21 +51,22 @@ auto thread_pool::resume(std::coroutine_handle<> handle) noexcept -> void
...
-    if (m_shutdown_requested.exchange(true, std::memory_order::acq_rel) == false)
+    if (m_shutdown_requested.exchange(true, std::memory_order::seq_cst) == false)
     {
         {
             // There is a race condition if we are not holding the lock with the executors
             // to always receive this.  std::jthread stop token works without this properly.
             std::unique_lock<std::mutex> lk{m_wait_mutex};
             m_wait_cv.notify_all();
         }
+        m_wait_cv.notify_all();
```
jbaldwin, line 69: This line should be unnecessary since its already called above on line 67

jbaldwin:
The failed test on fedora 32 is the ssl/tls bug I'm working through, so I retriggered that run.

jbaldwin (changes requested):

jbaldwin:
I also left just a few comments to really clean it up if you don't mind taking a look at them.

bruno-j-nicoletti:
I've removed the XML reporter and fixed the bits you requested. I can't see for the life of me why the precommit keeps doing weird thing the to readme.

jbaldwin:
There are a couple more `seq_cst` changes still, can you undo those across all the files please?

bruno-j-nicoletti:
Removed all the extraneous occurrences of seq_cst as requested.

Parallel programming is hard.

bruno-j-nicoletti:
22.04 dies on installing requirements via apt.

jbaldwin:
yeah, its a weird normal failure, I don't know why the package managers are so flaky but I see this all the time on lots of projects.  Only way to really get around it is either self hosting or baking the installs into a docker file (which I don't really want to maintain)

jbaldwin (approved):

bruno-j-nicoletti:
Hurrah!

PR #215, opened 2023-11-22
Make coro::io_scheduler using statements public

ripose-jp:
coro::io_scheduler::yield_until used a coro::io_scheduler::time_point as an arugment. While possible to use this method by passing a std::chrono::steady_clock::time_point, it is not clear from the API that this is the correct thing to do. Making coro::io_scheduler::time_point public allows it to be used as a type to construct time_points regardless of whether or not it may be redefined in the future.

Here's a minimal example to demonstrate the issue:
```cpp
#include <chrono>

#include <coro/coro.hpp>

int main(void)
{
    coro::io_scheduler io_scheduler;

    const auto yield_func = [&] () -> coro::task<int>
    {
        using namespace std::chrono_literals;
        co_await io_scheduler.yield_until(
            coro::io_scheduler::time_point{
                std::chrono::steady_clock::now().time_since_epoch() + 1s
            }
        );
        co_return 0;
    };

    return coro::sync_wait(yield_func());
}
```
While this is a bit of a contrived example, I don't believe it's unreasonable to expect it to compile.

### DIFF
--- a/include/coro/io_scheduler.hpp
+++ b/include/coro/io_scheduler.hpp
@@ -23,11 +23,11 @@ namespace coro
 {
 class io_scheduler
 {
+public:
     using clock        = detail::poll_info::clock;
     using time_point   = detail::poll_info::time_point;
     using timed_events = detail::poll_info::timed_events;
 
-public:
     class schedule_operation;
     friend schedule_operation;
 


jbaldwin:
I think its a good idea and you make a great point. I'm wondering if we should move it to a file like `coro/time.hpp` for the using statements to move it out of the `poll_info.hpp` file since its a details file and isn't normally supposed to be used/exposed to the end user. What do you think of that?

ripose-jp:
That works for me. My solution was just the easiest one, but its definitely not the best if it's not hiding types that shouldn't be user-facing. I'll close this since you probably have a better idea of how you want to implement this than I do.

PR #221, opened 2023-11-29
Use std::thread for coro::thread_pool

### LINKED ISSUE #220: coro:thread_pool use std::thread instead of std::jthread
This is in support of using emscripten, since that doesn't support std::jthread yet we'll use std::thread for now.

jbaldwin:
* This change switches from std::jthread to std::thread for the coro::thread_pool
* This change is in support of emscripten support

Closes #220

### DIFF
--- a/src/io_scheduler.cpp
+++ b/src/io_scheduler.cpp
@@ -183,13 +183,11 @@ auto io_scheduler::poll(fd_t fd, coro::poll_op op, std::chrono::milliseconds tim
 auto io_scheduler::shutdown() noexcept -> void
 {
     // Only allow shutdown to occur once.
-    if (m_shutdown_requested.exchange(true, std::memory_order::seq_cst) == false)
+    if (m_shutdown_requested.exchange(true, std::memory_order::acq_rel) == false)
     {
-        // std::cerr << "io_scheduler.m_shutdown_requested = " << m_shutdown_requested << "\n";
         if (m_thread_pool != nullptr)
         {
             m_thread_pool->shutdown();
-            // std::cerr << "io_scheduler.m_thread_pool->shutdown() complete\n";
         }
 
         // Signal the event loop to stop asap, triggering the event fd is safe.
--- a/src/io_scheduler.cpp
+++ b/src/io_scheduler.cpp
@@ -183,11 +183,13 @@ auto io_scheduler::poll(fd_t fd, coro::poll_op op, std::chrono::milliseconds tim
 auto io_scheduler::shutdown() noexcept -> void
 {
     // Only allow shutdown to occur once.
-    if (m_shutdown_requested.exchange(true, std::memory_order::acq_rel) == false)
+    if (m_shutdown_requested.exchange(true, std::memory_order::seq_cst) == false)
     {
+        // std::cerr << "io_scheduler.m_shutdown_requested = " << m_shutdown_requested << "\n";
         if (m_thread_pool != nullptr)
         {
             m_thread_pool->shutdown();
+            // std::cerr << "io_scheduler.m_thread_pool->shutdown() complete\n";
         }
 
         // Signal the event loop to stop asap, triggering the event fd is safe.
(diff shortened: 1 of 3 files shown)


src/io_scheduler.cpp
```diff
@@ -183,11 +183,13 @@ auto io_scheduler::poll(fd_t fd, coro::poll_op op, std::chrono::milliseconds tim
 auto io_scheduler::shutdown() noexcept -> void
 {
     // Only allow shutdown to occur once.
-    if (m_shutdown_requested.exchange(true, std::memory_order::acq_rel) == false)
+    if (m_shutdown_requested.exchange(true, std::memory_order::seq_cst) == false)
```
jbaldwin, line 186: switch back to `acquire`

src/io_scheduler.cpp
```diff
@@ -183,11 +183,13 @@ auto io_scheduler::poll(fd_t fd, coro::poll_op op, std::chrono::milliseconds tim
...
     {
+        // std::cerr << "io_scheduler.m_shutdown_requested = " << m_shutdown_requested << "\n";
```
jbaldwin, line 188: remove

src/io_scheduler.cpp
```diff
@@ -183,11 +183,13 @@ auto io_scheduler::poll(fd_t fd, coro::poll_op op, std::chrono::milliseconds tim
...
         if (m_thread_pool != nullptr)
         {
             m_thread_pool->shutdown();
+            // std::cerr << "io_scheduler.m_thread_pool->shutdown() complete\n";
```
jbaldwin, line 192: remove

src/thread_pool.cpp
```diff
@@ -75,34 +77,51 @@ auto thread_pool::shutdown() noexcept -> void
...
+            // Release the lock while executing the coroutine.
+            lk.unlock();
             handle.resume();
             m_size.fetch_sub(1, std::memory_order::release);
+            lk.lock();
         }
+    };
+
+    // Process until shutdown is requested and the total number of tasks reaches zero.
+    while (!m_shutdown_requested.load(std::memory_order::acquire))
```
jbaldwin, line 104: todo: combine these, I don't think the `wait_for` on the cv is required if we find the bug

src/io_scheduler.cpp
jbaldwin: `acq_rel`***

jbaldwin:
@bruno-j-nicoletti I think I found it!  Its the `notify_all()` in the shutdown, its got some kind of a race condition and by placing the lock around it before calling it guarantees each worker thread in the `coro::thread_pool` will properly catch it.  Clearly the `std::jthread` `stop_token` handled this properly :grimacing: .  Please take a look, I did find one `relaxed` instance that was probably wrong so that got fixed as well.

src/io_scheduler.cpp
```diff
@@ -253,11 +253,10 @@ auto io_scheduler::process_events_execute(std::chrono::milliseconds timeout) ->
                 // Process scheduled coroutines.
                 process_scheduled_execute_inline();
             }
-            else if (handle_ptr == m_shutdown_ptr)
-                [[unlikely]]
-                {
-                    // Nothing to do , just needed to wake-up and smell the flowers
-                }
+            else if (handle_ptr == m_shutdown_ptr) [[unlikely]]
```
jbaldwin, line 256: this line got reformatted due to other changes that I backed out

src/thread_pool.cpp
```diff
@@ -35,7 +35,7 @@ thread_pool::~thread_pool()
 
 auto thread_pool::schedule() -> operation
 {
-    if (!m_shutdown_requested.load(std::memory_order::relaxed))
+    if (!m_shutdown_requested.load(std::memory_order::acquire))
```
jbaldwin, line 38: This shouldn't have been `relaxed` since we want to `acquire` any other writes from other threads.

src/thread_pool.cpp
```diff
@@ -75,33 +77,35 @@ auto thread_pool::shutdown() noexcept -> void
...
+        std::unique_lock<std::mutex> lk{m_wait_mutex};
+        m_wait_cv.wait(
+            lk,
+            [&] {
+                return m_size.load(std::memory_order::acquire) > 0 ||
+                       m_shutdown_requested.load(std::memory_order::acquire);
+            });
+
+        // Process this batch until the queue is empty.
+        while (!m_queue.empty())
```
jbaldwin, line 99: I've switched the logic here to drain the queue without the condition variable being invoked each time between executions of tasks

bruno-j-nicoletti (approved):

bruno-j-nicoletti (approved):

PR #234, opened 2024-01-08
CMake: Keep git hooks as optional

uilianries:
When not developing the project, but only building on CI or packaging, using git config for hooks is not needed and could be skipped. However, there is no such option for it, so users need to patch the cmake file to avoid.

This PR introduces the option `LIBCORO_RUN_GITCONFIG` (default ON) to execute or not git config. 

It should not change the current behavior, but it's useful for who wants to build the project only.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,17 +12,20 @@ if (NOT "$ENV{version}" STREQUAL "")
     set(PROJECT_VERSION "$ENV{version}" CACHE INTERNAL "Copied from environment variable")
 endif()
 
-# Set the githooks directory to auto format and update the readme.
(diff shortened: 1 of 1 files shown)


jbaldwin:
Maybe the default should actually be off? Most people will be building to use it, not develop?

uilianries:
> Maybe the default should actually be off? Most people will be building to use it, not develop?

I kept ON to use the same behavior, but yes. Plus, using [cmake-presents](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html), is possible to define a "profile" with it enabled by default, so don't need to pass =ON when developing, for each build. 

I'll update it to OFF. Thank you for your quick review.

uilianries:
Done.

jbaldwin (approved):

PR #240, opened 2024-01-18
Fix use after free in coro::task_container

ripose-jp:
In coro::task_container, this is potential for a use after free bug to occur between the final co_return of make_cleanup_task and another thread executing gc_internal. This can happen because there is a delay between the std::scoped_lock being destroyed upon the co_return of make_cleanup_task and its final suspend. This leads to a race condition where another thread can acquire the lock and initiate a garbage collection, where that cleanup task is destroyed before it's final suspend. This leads to a use after free.

This commit fixes the issue by reworking how tasks get garbage collected. Instead of unconditionally destroying tasks scheduled for deletion, the gc_internal calls is_ready on the coro::task first. This function will only be true in the cases where the task has entered final suspend, has been default constructed, or has been destroyed. This means a task can get scheduled for deletion, but not be deleted when gc_internal is called. This allows time for the task to enter final suspend, which avoids the use after free bug.

To make this implementation easier, several aspects of the implementation were changed. m_task_indexes was removed in favor of a queue which keeps track of free indices into m_tasks. This makes things easier since it removes the hard constraint of keeping used indices behind m_free_pos and free indices at or after. m_tasks_to_delete was then changed to a list in order to allow for constant time deletion of elements in place. This is helpful because it is possible for tasks to be queue for deletion, but not yet safe to delete. This means unconditionally clearing the m_tasks_to_delete data structure is no longer viable.

---

Here's some example code that can cause the use-after-free bug. Since it's a race condition, it's non-deterministic so it may take some time before the bug occurs.

```cpp
#include <coro/coro.hpp>

#include <cstdint>
#include <iostream>
#include <numeric>
#include <vector>

static constexpr size_t DUMMY_TASK_COUNT{10};

static coro::task<void> do_dummy_work()
{
    constexpr size_t SIZE_DOUBLE_WORDS{100 * 1024 * 1024 / sizeof(uint64_t)};

    [[maybe_unused]] std::vector<uint64_t> buffer;
    buffer.resize(SIZE_DOUBLE_WORDS);
    std::iota(std::begin(buffer), std::end(buffer), 0);
    co_return;
}

static coro::task<void> run_async(
    std::shared_ptr<coro::thread_pool> &tp,
    coro::task_container<coro::thread_pool> &tc)
{
    co_await tp->schedule();

    while (true)
    {
        for (size_t i{0}; i < DUMMY_TASK_COUNT; ++i)
        {
            tc.start(do_dummy_work());
        }
        co_await tc.garbage_collect_and_yield_until_empty();
        std::cout << "Dummy work iteration complete\n";
    }
}

int main(void)
{
    std::shared_ptr<coro::thread_pool> tp = std::make_shared<coro::thread_pool>(
        coro::thread_pool::options{ .thread_count = DUMMY_TASK_COUNT, }
    );
    coro::task_container tc{tp};
    coro::sync_wait(run_async(tp, tc));
    return 0;
}
```

### DIFF
--- a/include/coro/task_container.hpp
+++ b/include/coro/task_container.hpp
@@ -9,6 +9,7 @@
 #include <list>
 #include <memory>
 #include <mutex>
+#include <queue>
 #include <vector>
 
 namespace coro
@@ -19,8 +20,6 @@ template<concepts::executor executor_type>
 class task_container
 {
 public:
-    using task_position = std::list<std::size_t>::iterator;
-
     struct options
     {
         /// The number of task spots to reserve space for upon creating the container.
@@ -87,17 +86,17 @@ public:
         }
 
         // Only grow if completely full and attempting to add more.
-        if (m_free_pos == m_task_indexes.end())
+        if (m_free_task_indices.empty())
         {
-            m_free_pos = grow();
+            grow();
         }
 
-        // Store the task inside a cleanup task for self deletion.
-        auto index     = *m_free_pos;
-        m_tasks[index] = make_cleanup_task(std::move(user_task), m_free_pos);
+        // Reserve a free task index
+        std::size_t index = m_free_task_indices.front();
+        m_free_task_indices.pop();
 
-        // Mark the current used slot as used.
-        std::advance(m_free_pos, 1);
+        // Store the task inside a cleanup task for self deletion.
+        m_tasks[index] = make_cleanup_task(std::move(user_task), index);
 
         // Start executing from the cleanup task to schedule the user's task onto the thread pool.
         m_tasks[index].resume();
@@ -112,24 +111,6 @@ public:
     {
         std::scoped_lock lk{m_mutex};
         return gc_internal();
-    }
-
-    /**
-     * @return The number of tasks that are awaiting deletion.
-     */
-    auto delete_task_size() const -> std::size_t
-    {
-        std::atomic_thread_fence(std::memory_order::acquire);
-        return m_tasks_to_delete.size();
-    }
-
-    /**
-     * @return True if there are no tasks awaiting deletion.
-     */
-    auto delete_tasks_empty() const -> bool
-    {
-        std::atomic_thread_fence(std::memory_order::acquire);
-        return m_tasks_to_delete.empty();
     }
 
     /**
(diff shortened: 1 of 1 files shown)


jbaldwin:
I'll take a deep dive into this as soon as I can, the description of the use after free makes sense, and thank you for pushing a detailed PR with a solution to the problem.

include/coro/task_container.hpp
```diff
@@ -172,38 +153,42 @@ class task_container
...
+                continue;
             }
-            deleted = m_tasks_to_delete.size();
-            m_tasks_to_delete.clear();
+            // Destroy the cleanup task and the user task.
+            m_tasks[*pos].destroy();
+            // Put the deleted position at the end of the free indexes list.
+            m_free_task_indices.emplace(*pos);
+            // Remove index from tasks to delete
+            m_tasks_to_delete.erase(pos++);
```
jbaldwin, line 187: From https://en.cppreference.com/w/cpp/container/deque/erase (std::queue is a specialization on std::deque)

```
std::deque<T,Allocator>::erase
...
Complexity

Linear: the number of calls to the destructor of T is the same as the number of elements erased, the number of calls to the assignment operator of T is no more than the lesser of the number of elements before the erased elements and the number of elements after the erased elements. 
```

I think the second part of this is important in that its `O(n)` in elements being shifted from where the element was erased, so an item being erased in the middle is shifting n/2 items I think? The vector solution with swapping items doesn't have this linear problem which is why I used it to make sure this function doesn't degrade worst case into `O(n^2). Which I think in theory is possible if it keeps having to shift half the elements on each delete.

jbaldwin (changes requested):

include/coro/task_container.hpp
ripose-jp: `m_tasks_to_delete` is a `std::list` which has constant time `erase`.

jbaldwin: Oops, I reviewed this too fast yesterday and mixed the two variables up :sweat_smile:, my bad. Thanks for pointing that out.

include/coro/task_container.hpp
```diff
@@ -172,38 +153,42 @@ class task_container
...
+            // Indicate a task was deleted.
+            ++deleted;
         }
+        m_size.fetch_sub(deleted, std::memory_order::relaxed);
```
jbaldwin, line 191: I'm concerned about this change as well, the user task is "done" when `make_cleanup_task` adds itself to the `m_tasks_to_delete` list. This behavior will now make it seem like those tasks are still alive and running and won't drop until the user adds a new task and gc is invoked, or they manually invoke gc. Doesn't it still make sense to reduce the count at the end of `make_cleanup_task` to properly reflect how many user tasks are alive and running?

include/coro/task_container.hpp
jbaldwin: I agree on this solution, since we cannot unconditionally "clear" the deleted tasks and reuse that memory a std::list for tasks ready to delete is an appropriate data structure. I might run some performance tests and see if the allocations make a regression on the performance (it probably will?), in which case we can reintroduce pre-allocating nodes up to the growth size and re-using them by swapping, etc. I'll leave that off this PR though since we are fixing a bigger bug at the moment.  The free task index list can probably re-use this strategy as well to reduce allocations.

jbaldwin (changes requested):

include/coro/task_container.hpp
jbaldwin: I'll close this comment out since I just misunderstood the original implementation.

include/coro/task_container.hpp
ripose-jp: I believe that would create a bug in the destructor and `garbage_collect_and_yield_until_empty` since both of those methods rely on calling `empty` to determine if all tasks have been garbage collected. It's possible that no tasks end up getting garbage collected when calling these methods since it would be possible for `m_size` to be 0 before `gc_internal` ever runs.

If the current behavior of `size` and `empty` is a concern, I could create a private size that tracks non-garbage collected tasks and running tasks. Program logic could use this without interfering with the public facing size that tracks only running tasks.

jbaldwin: You're absolutely right. I've been thinking on it and wondering if its worth it to track? Live vs Used?

ripose-jp: For my use case it's not helpful. I could image it might be useful for profiling or statistical purposes, but I don't know what else. I don't have a strong opinion on the matter, so I'll leave it to your judgement.

jbaldwin: The user's task is cleaned up so resources like sockets are free'ed, so you're right its probably not that useful. Thanks for talking through it, I'll get this merged.

jbaldwin (approved):

PR #246, opened 2024-02-08
Fix for higher version of C++ usage, Clang support, and some build policy change.

stripe2933:
1. I removed the hardcoding in the CMake script that enables C++20 mode, aiming to use compiler versions that support coroutines. However, it now forcefully uses C++20 mode even in higher versions (e.g., C++23, C++26). For instance, a C++23 project utilizing this library won't be able to use `std::print`.
2. Although apps using this project as a dependency do not require test/example builds, the build defaults for these are set to true. The new commit activates the build only when the project is at the top level.
3. Despite the inability to use Linux kernel functions on macOS, the corresponding build flag is enabled unless the build environment is Linux and not WebAssembly. I modified it to activate the flag only in a Linux environment that is not WebAssembly.
4. In the libc++ environment, `exception_ptr` now requires inclusion of the `<exception>` header.


jbaldwin:
Hi Thanks for the PR, these look like great additions.

(1) I agree hardcoding to C++20 was a mistake, it was initially there a few years ago when it was the latest standard but it is definitely a problem now since C++23 and onward are coming out.

(2) If I'm understanding this correctly you can now only build tests and examples if you build `libcoro` directly from the root checkout directory, and not if its included in another project as a dependency? I'm struggling to think of why you'd need to build tests and examples in this scenario but the change does prevent you from doing so if I'm reading it correctly.

(3) :+1: 

(4) Great, looks like it was also reported in #247 (@bruno-j-nicoletti looks like this PR fixes your build problem I think)

stripe2933:
(2): In my case, I tried to use vcpkg for 1st tier package management, but it failed to build. So I tried second attempt with using [`CPMAddPackage`](https://github.com/cpm-cmake/CPM.cmake). It directly clone the repository or release archive into project build dir, but in this case I should build libcoro's tests and examples, which is not needed for library use. Therefore, I think it is feasible to build them only if project is master (top level).

jbaldwin:
Yeah cool, it makes sense from that perspective. I wonder if it's worth adding CI runs for the package managers to download and link it to a sample project just to know it's working properly. I'll open an issue to possibly take a look at the major ones.

jbaldwin (approved):

PR #264, opened 2024-05-23
Add macos 14

jbaldwin:
The latest macos image appears to have changed the directory that llvm/clang are installed into. Change the directory appropriately.


jbaldwin:
This was fixed by #318

PR #265, opened 2024-05-24
coro::thread_pool high cpu usage when tasks < threads

### LINKED ISSUE #262: High CPU usage, when thread_count is greater than the number of tasks
```c++
#include <coro/coro.hpp>

coro::task<int> sleep_for(int seconds) {
    std::this_thread::sleep_for(std::chrono::seconds {seconds});
    co_return 0;
}

coro::task<> wait_for_task(auto &Pool, int delaySeconds) {
    co_await Pool.schedule();
    while (true) {
        co_await sleep_for(delaySeconds);
        std::println("[{}]wait for {} seconds", std::chrono::system_clock::now(), delaySeconds);
    }
    co_return;
}

int main() {
    // cpu high when 'thread_count >= 3'
    coro::thread_pool pool {coro::thread_pool::options {.thread_count = 3}};
    coro::sync_wait(coro::when_all(wait_for_task(pool, 1), wait_for_task(pool, 3)));

    return 0;
}
```

When two tasks are scheduled to any two threads, the extra threads will occupy the CPU of the entire CPU thread.

jbaldwin:
The check for m_size > 0 was keeping threads awake in a spin state until all tasks completed. This correctl now uses m_queue.size() behind the lock to correctly only wake up threads on the condition variable when tasks are waiting to be processed.

Closes #262

### DIFF
--- a/src/thread_pool.cpp
+++ b/src/thread_pool.cpp
@@ -44,15 +44,21 @@ auto thread_pool::schedule() -> operation
     throw std::runtime_error("coro::thread_pool is shutting down, unable to schedule new tasks.");
 }
 
-auto thread_pool::resume(std::coroutine_handle<> handle) noexcept -> void
+auto thread_pool::resume(std::coroutine_handle<> handle) noexcept -> bool
 {
     if (handle == nullptr)
     {
-        return;
+        return false;
+    }
+
+    if (m_shutdown_requested.load(std::memory_order::acquire))
+    {
+        return false;
     }
 
     m_size.fetch_add(1, std::memory_order::release);
     schedule_impl(handle);
+    return true;
 }
 
 auto thread_pool::shutdown() noexcept -> void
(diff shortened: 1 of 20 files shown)


src/thread_pool.cpp
```diff
@@ -44,28 +44,24 @@ auto thread_pool::schedule() -> operation
...
 
 auto thread_pool::shutdown() noexcept -> void
 {
     // Only allow shutdown to occur once.
     if (m_shutdown_requested.exchange(true, std::memory_order::acq_rel) == false)
     {
-        {
-            // There is a race condition if we are not holding the lock with the executors
-            // to always receive this.  std::jthread stop token works without this properly.
-            std::unique_lock<std::mutex> lk{m_wait_mutex};
```
jbaldwin, line 66: re-add, shouldn't make everything wait 1000ms to quit

PR #272, opened 2024-07-11
Use lock for sync_wait completion

### LINKED ISSUE #270: `coro::sync_wait(coro::when_all())` paradigm (very rarely) hangs
When running a set of tasks in parallel to accomplish a task, to assure completion, one typically uses a `coro::sync_wait(coro::when_all(std::move(tasks));` line. Sometimes, but very rarely, this hangs at the https://github.com/jbaldwin/libcoro/blob/d4a55159f9a61509dd30668672b5374147912bb0/src/sync_wait.cpp#L23 line.

Curiously, `m_set` is true when checked in the debugger - I'm not sure how that can be the case when that line is stuck waiting. The code looks proper and I suspect some subtle race condition. As an aside, since this is C++20, we don't need to use the condition variable because atomics have [atomic::wait()](https://en.cppreference.com/w/cpp/atomic/atomic/wait) and [atomic::notify_all()](https://en.cppreference.com/w/cpp/atomic/atomic/notify_all) and [atomic::notify_one()](https://en.cppreference.com/w/cpp/atomic/atomic/notify_one) - I'm not sure if refactoring to use those will fix the issue or not (I don't currently have time to try).

I've taken to doing something like:
```c++
auto when_all = [](coro::thread_pool &my_thread_pool, std::vector<coro::task<void>>&& my_tasks) -> coro::task<void>
{
   co_await tp.schedule();
   co_await coro::when_all(std::move(my_tasks));
};
coro::sync_wait(when_all(thread_pool, std::move(tasks)));
```
which seems to mitigate the issue but also, in my case, adds a noticeable performance hit (about 1-5%). I just tried this as a wild guess that `coro::when_all` needed to slow down a bit by going on the thread pool.

The hang does seem more prevalent when `tasks.size()` is higher. That is, when processing only 1-5 tasks, I'm not sure I've seen it happen. When it is 50-100 tasks, it seem to happen more regularly (like one out of every 80,000 or so calls). In my case, the tasks are very short lived (like 10 lookups in a `std::map()`).

I've diagnosed this on a 16-core box with source compiled with Microsoft Visual C++ in debug mode. I _believe_ I've seen the same behavior on Linux compiled with g++13 but I didn't run in the debugger there so I cannot be certain. I had built in both debug and release on Linux and I don't recall which executable I was running when I saw the issue.

jbaldwin:
* release/acquire memory ordering has a race condition
* also reproduced on seq_cst

Closes #270


test/main.cpp
```diff
@@ -16,19 +16,20 @@ struct test_setup_networking
         signal(SIGPIPE, SIG_IGN);
 
     #ifdef LIBCORO_FEATURE_TLS
-        // For SSL/TLS tests create a localhost cert.pem and key.pem, tests expected these files
-        // to be generated into the same directory that the tests are running in.
-        auto unused = system(
-            "openssl req -x509 -newkey rsa:4096 -keyout key.pem -out cert.pem -days 365 -subj '/CN=localhost' -nodes");
-        (void)unused;
+            // For SSL/TLS tests create a localhost cert.pem and key.pem, tests expected these files
```
jbaldwin, line 19: remove before merging

PR #288, opened 2025-01-29
coro::task_container gc fix not completing coroutines

### LINKED ISSUE #287: schedule stress test
Hi [Josh Baldwin],

I have a question about the behavior of the StressTest function. I expected g_count to be exactly 200,000 after all tasks are scheduled and executed. However, in practice, the value of g_count is less than 200,000.

Could you clarify why this happens? Is it due to tasks being dropped, preemption, or some other issue in the scheduler or coroutine execution?

Thank you for your insight!
```C++
std::atomic_uint32_t g_count = 0;
static void StressTest() {
  auto scheduler = coro::io_scheduler::make_shared(
      coro::io_scheduler::options{.pool = coro::thread_pool::options{.thread_count = 1}});

  auto task1 = [&]() -> coro::task<void> {
    g_count++;
    co_return;
  };

  for (int i = 0; i < 200000; ++i) {
    scheduler->schedule(task1());
  }

  std::this_thread::sleep_for(std::chrono::seconds(10));
  spdlog::info("g_count = {}", g_count.load());
}
```

jbaldwin:
The coro::task_container::gc_internal function was deleting coroutines when marked as .done(), however there is some mechanism that rarely would cause the user_task coroutine to not actually execute. I'm still not sure exactly why this is the case but:

1) Simply disabling gc_internal() would stop the problem 
2) Running gc_internal() and moving the coro::task to a 'dead' list
   still caused the issue.

With these in mind I spent time re-reading the specification on the final_suspend and determined that coro::task_container should be a thing atomic counter to track the submitted coroutines and have them self delete. The self deletion is now done via a
coro::detail::task_self_destroying coroutine type that takes advantage of the promise's final_suspend() not suspending. The spec states that if this doesn't suspend then the coroutine will call destroy() on itself.

Closes #287


include/coro/detail/task_self_destroying.hpp
```diff
@@ -0,0 +1,66 @@
...
+    auto operator=(const promise_self_destroying&) -> promise_self_destroying& = delete;
+    auto operator=(promise_self_destroying&&) -> promise_self_destroying&;
+
+    auto get_return_object() -> task_self_destroying;
+    auto initial_suspend() -> std::suspend_always;
+    auto final_suspend() noexcept -> std::suspend_never;
+    auto return_void() noexcept -> void;
+    auto unhandled_exception() -> void;
+
+    auto task_container_size(std::atomic<std::size_t>& task_container_size) -> void;
```
breakds, line 29: Small suggestion: would `set_task_container_size_ref` sound more expressive?

include/coro/io_scheduler.hpp
```diff
@@ -178,7 +178,7 @@ class io_scheduler : public std::enable_shared_from_this<io_scheduler>
      * @param task The task to execute on this io_scheduler.  It's lifetime ownership will be transferred
      *             to this io_scheduler.
      */
-    auto schedule(coro::task<void>&& task) -> void;
+    auto schedule(coro::task<void>&& task) -> bool;
```
breakds, line 181: Suggestion: Also add to the documentation above what the return value means?

breakds (approved):
With this approach the logic is actually a lot simpler. Thanks!

include/coro/io_scheduler.hpp
jbaldwin: Good idea, I'll also add docs for the `task_container::start` function as well.

include/coro/detail/task_self_destroying.hpp
jbaldwin: I could have sworn I read that the [https://isocpp.github.io/CppCoreGuidelines/](CppCoreGuidelines) suggested not having `get` and `set` prefixes on methods because its obvious that getters have a return type and setters take an argument and I've pretty much kept to that for this library. However I'm unable to find where this is, or if it ever was a thing... but you'll notice on most of the classes that is the style I've kept for this project.

e.g. `coro::task<>::handle() -> std::coroutine_handle<>` gets the underlying coroutine handle, but isn't named `get_handle()`.

I think I'd prefer to keep the consistency with how large this project is, but I appreciate the comment, hopefully the reasoning makes sense.

jbaldwin:
I renamed `task_self_destroying` to `task_self_deleting` since its more inline with the `delete` keyword.

include/coro/detail/task_self_destroying.hpp
breakds: Sounds good to me.

breakds (approved):

PR #304, opened 2025-02-22
Feature/condition variable compiler fixes

jbaldwin:


include/coro/when_any.hpp
```diff
@@ -23,16 +24,26 @@ namespace coro
...
     coro::mutex&                m,
     std::atomic<bool>&          return_value_set,
     coro::event&                notify,
     std::optional<return_type>& return_value,
     awaitable                   a) -> coro::task<void>
 {
-    auto result      = co_await static_cast<awaitable&&>(a);
-    auto scoped_lock = co_await m.lock();
+    auto result = co_await static_cast<awaitable&&>(a);
+    co_await m.lock();
```
dobord, line 44: maybe there is a missing `auto scoped_lock = ` here

include/coro/when_any.hpp
```diff
@@ -23,16 +24,26 @@ namespace coro
...
+    // Its important to only touch return_value and notify once since their lifetimes will be destroyed
+    // after being set ane notified the first time.
     if (return_value_set.load(std::memory_order::acquire) == false)
     {
         return_value_set.store(true, std::memory_order::release);
```
dobord, line 49: To improve the atomicity of the operation of checking `return_value_set `for `false `and setting it to `true`, it would be great to use compare_exchange here. For example, like this:
```
auto expected_value = false;
if (return_value_set.compare_exchange_strong(expected_value, true, std::memory_order::acq_rel))
{

```

PR #311, opened 2025-03-25
Adds coro::queue<T>

### LINKED ISSUE #307: Add new object coro::queue<T>
This structure should be coroutine safe and be backed by a growable queue, making it different from the statically sized coro::ring_buffer.

The API should probably be as similar to `std::queue` as possible but with async methods.

jbaldwin:
* Pushing elements into the queue will either wake a waiter to immediately consume the value, or push into the queue if there are no current waiters.
* Popping elements from the queue will either consume the front immediately if there are elements available, or wait for an element to be pushed.

This structure differs from `coro::ring_buffer` in that it has an unbounded size and also means that `push()` calls are not coroutines since instead of waiting for a slot to open will simply allocate memory to add the item. Only the `pop()` calls can be awaited and will suspend the coroutine until an item is available.

Use `coro::queue::notify_waiters()` to shutdown the queue and wake all waiters.

Closes #307

### DIFF
--- a/test/test_queue.cpp
+++ b/test/test_queue.cpp
@@ -16,8 +16,8 @@ TEST_CASE("queue shutdown produce", "[queue]")
         co_return std::move(*expected);
     };
 
-    q.shutdown_notify_waiters();
-    q.push(42);
+    coro::sync_wait(q.shutdown_notify_waiters());
+    coro::sync_wait(q.push(42));
 
     auto result = coro::sync_wait(make_consumer_task(q));
     REQUIRE(result == 0);
@@ -38,7 +38,7 @@ TEST_CASE("queue single produce consume", "[queue]")
         co_return std::move(*expected);
     };
 
-    q.push(42);
+    coro::sync_wait(q.push(42));
 
     auto result = coro::sync_wait(make_consumer_task(q));
     REQUIRE(result == 42);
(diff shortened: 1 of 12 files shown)


test/test_queue.cpp
```diff
@@ -0,0 +1,211 @@
...
+    }
+}
+
+TEST_CASE("queue produce consume direct", "[queue]")
+{
+    const uint64_t        ITERATIONS = 10;
+    coro::queue<uint64_t> q{};
+    coro::thread_pool     tp{};
+
+    auto make_producer_task = [&ITERATIONS](coro::thread_pool& tp, coro::queue<uint64_t>& q) -> coro::task<uint64_t>
```
dobord, line 83: There is no need to capture the ITERATIONS constant here. The compiler hints about it here and in several similar places further down the code.

> warning: lambda capture 'ITERATIONS' is not required to be captured for this use [-Wunused-lambda-capture]
>    83 |     auto make_producer_task = [&ITERATIONS](coro::thread_pool& tp, coro::queue<uint64_t>& q) -> coro::task<uint64_t>

dobord (commented):
This is a cool feature! I checked the tests with ASAN, TSAN, USAN - everything goes without any complaints

test/test_queue.cpp
jbaldwin: Cool I'll remove it!

jbaldwin:
I've been on vacation, hence my absence, hoping to get back to these PRs soon.

jbaldwin:
@dobord I made some adjustments to try and make the api a little more async friendly, and also fixed the comment you had posted.

Issue #315, opened 2025-05-03
Add kqueue backend to support networking on macOS and BSD


tglane:
Currently only `epoll` is used as the backend in the `io_scheduler` so the IO only works on linux at the current state. By adding an additional IO backend for `kqueue` we could also support macOS and I think BSD systems as well without changing other parts of the library. The timers need to be replaced under macOS as well because there is no `timer_fd` support right now.

I already experimented with that in a fork [here](https://github.com/tglane/libcoro/pull/1) so you might take a look. If this is a desired feature for this project I could clean up my fork and create a PR for further discussions on implementation details.

jbaldwin:
Hi @tglane I have been thinking on doing this for years so yes I'm absolutely interested in this change for the library. In the past I've also considered just using something like libuv for cross platform io event support but so far have decided against it to not bring in the dependency.

As an fyi I currently only have a Linux and Windows box for testing on and have an issue open to get the Mac CI working again, but I was unable to blindly debug my way through it so if you have any insight on that as well it would be appreciated.

I'm curious to see how your branch looks for kqueue!

tglane:
Ok that sound good. I'd be very interested in implementing this (implementing it yourself is more fun than using libuv anyway ;)). I might create a PR later so that we can discuss the details there?

I saw your branch regarding the macOS CI problems. However I think I got it to work with some minor adjustments because the whole CI pipeline completes even with the macOS tests on the branch I created and linked in the issue. There is just one test for `coro::when_any` that seems flaky and needs some investigation.

jbaldwin:
Great I'll keep an eye out for the branch!

I remember on the Mac build, it's that GitHub is deprecating the runner version the current CI is using so it needs to be migrated but apple seems to have moved where the compilers get installed and I couldn't get the path right from debugging within the action 😂😭

jbaldwin:
```bash
1: ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1: libcoro_test is a Catch2 v3.3.2 host application.
1: Run with -? for options
1: 
1: -------------------------------------------------------------------------------
1: when_any tuple return void (monostate)
1: -------------------------------------------------------------------------------
1: ../test/test_when_any.cpp:45
1: ...............................................................................
1: 
1: ../test/test_when_any.cpp:95: FAILED:
[...]

PR #317, opened 2025-05-05
feat: add kqueue support

tglane:
As discussed in #315 this PR attempts to add networking support for macOS (and perhaps BSD systems as well) by refactoring the asynchronous IO system to utilize an `io_notifier` to abstract away the linux-only `epoll` system. For macOS the `io_notifier` uses `kqueue` as the internal backend.


tglane:
@jbaldwin Could you trigge the remaining CI tests that are currently blocked?

jbaldwin:
If you want to put the macos ci into its own PR we can merge that quickly once it passes, then this PR will auto let you run the jobs once you contributed iirc

tglane:
> If you want to put the macos ci into its own PR we can merge that quickly once it passes, then this PR will auto let you run the jobs once you contributed iirc

Sounds like a good idea. I will move the CI changes into a separate PR and submit that, too.

jbaldwin:
I've done an initial pass and I've got a good grasp on what's going on. I'd like to spend more time on a deep drive review though, just wanted to let you know I am reviewing.

test/bench.cpp
```diff
@@ -534,7 +540,9 @@ TEST_CASE("benchmark tcp::server echo server thread pool", "[benchmark]")
         std::cerr << ms.count() << " : " << count << "\n";
     }
 }
+#endif
 
+/*
```
jbaldwin, line 545: Can we re-enable these tests?

test/main.cpp
```diff
@@ -26,9 +26,11 @@ struct test_setup_networking
 
     ~test_setup_networking()
     {
+    #ifdef LIBCORO_FEATURE_TLS
```
jbaldwin, line 29: :+1:

src/io_notifier.cpp
```diff
@@ -0,0 +1,256 @@
...
+    if (m_fd != -1)
+    {
+        ::close(m_fd);
+        m_fd = -1;
+    }
+}
+
+timer_handle::timer_handle(const void* timer_handle_ptr, io_notifier& notifier)
+#if defined(__FreeBSD__) || defined(__APPLE__) || defined(__OpenBSD__) || defined(__NetBSD__)
+    : m_fd{1},
```
jbaldwin, line 29: I'm not sure I understand how this works having a hardcoded file descriptor, can you explain whats going on here? I can see its used below on `watch_timer()`

src/io_notifier.cpp
```diff
@@ -0,0 +1,256 @@
...
+#if defined(__FreeBSD__) || defined(__APPLE__) || defined(__OpenBSD__) || defined(__NetBSD__)
+    (void)notifier;
+#elif defined(__linux__)
+    notifier.watch(m_fd, coro::poll_op::read, const_cast<void*>(m_timer_handle_ptr), true);
+#endif
+}
+
+auto io_notifier::watch_timer(const timer_handle& timer, std::chrono::nanoseconds duration) -> bool
+{
+#if defined(__FreeBSD__) || defined(__APPLE__) || defined(__OpenBSD__) || defined(__NetBSD__)
```
jbaldwin, line 44: its possible `duration` has a negative value here, can you handle that just like the linux branch does?

src/io_notifier.cpp
```diff
@@ -0,0 +1,256 @@
+#include "coro/io_notifier.hpp"
```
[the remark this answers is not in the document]
tglane: Another possible way would be to have a `io_notifier.hpp` in which we include one of `io_notifer_{epoll | kqueue}.hpp` depending on the ifdef macro. That way we would only have the ugly macro one time and the included plattform-depending header would be nice to read without any macros? I did it that way in a private project [here](https://github.com/tglane/socketwrapper/blob/v2/include/socketwrapper/detail/event_notifier.hpp). What do you think?

src/io_notifier.cpp
```diff
@@ -0,0 +1,256 @@
...
+
+#include <array>
+#include <stdexcept>
+
+namespace coro
+{
+
+io_notifier::io_notifier()
+#if defined(__FreeBSD__) || defined(__APPLE__) || defined(__OpenBSD__) || defined(__NetBSD__)
```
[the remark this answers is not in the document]
tglane: I'm not a big fan of that as it hides the details and includes another lookup step in order to understand what is going on. I think your request to move the implementations of the two implementations of `io_notifier` into separate files would be a better way to deal with that issue here.

include/coro/io_scheduler.hpp
```diff
@@ -365,8 +365,8 @@ class io_scheduler : public std::enable_shared_from_this<io_scheduler>
             if (m_schedule_fd_triggered.compare_exchange_strong(
                     expected, true, std::memory_order::release, std::memory_order::relaxed))
             {
-                eventfd_t value{1};
-                eventfd_write(m_schedule_fd, value);
+                const int stop = 1;
```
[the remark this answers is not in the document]
tglane: No, I guess it was just a copy-paste error. I will change it back to `value`.

src/io_notifier.cpp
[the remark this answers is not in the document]
jbaldwin: I think this would be a very reasonable approach.

src/io_notifier.cpp
tglane: Yeah that is a bit counterintuitive with kqueue. Kqueue directly supports timers. But it is not using real file descriptors. The `fd` parameter for the kqueue timer is just an arbitrary identifier used by the kqueue code to identify the very timer you want to handle. So we can use any integer we want. With that being said, I think we might want to rework this to use a random number instead of a fixed one since it theoretically is possible to create multiple instances of `timer_handle` outside of our library code.

See kqueue documentation to the timer filter:
```
EVFILT_TIMER	   Establishes an arbitrary timer identified by	ident.
			   When	adding a timer,	data specifies the  moment  to
			   fire	 the  timer  (for NOTE_ABSTIME)	or the timeout
			   period.   The  timer	 will	be   periodic	unless
			   EV_ONESHOT  or  NOTE_ABSTIME	 is specified.	On re-
			   turn, data contains the number of times the timeout
			   has expired since the last call to  kevent().   For
			   non-monotonic  timers,  this	 filter	 automatically
			   sets	the EV_CLEAR flag internally.

			   The filter  accepts	the  following	flags  in  the
			   fflags argument:

			   NOTE_SECONDS	     data is in	seconds.

			   NOTE_MSECONDS     data is in	milliseconds.

			   NOTE_USECONDS     data is in	microseconds.

			   NOTE_NSECONDS     data is in	nanoseconds.

			   NOTE_ABSTIME	     The  specified expiration time is
					     absolute.

			   If fflags is	not set, the default is	 milliseconds.
			   On  return,	fflags contains	the events which trig-
			   gered the filter.

			   Periodic timers with	a specified timeout of 0  will
			   be silently adjusted	to timeout after 1 of the time
			   units  specified  by	 the  requested	 precision  in
			   fflags.  If an absolute time	is specified that  has
			   already  passed,  then it is	treated	as if the cur-
			   rent	time were specified and	the event will fire as
			   soon	as possible.

			   If an existing  timer  is  re-added,	 the  existing
			   timer  will	be effectively canceled	(throwing away
			   any undelivered record of  previous	timer  expira-
			   tion)  and re-started using the new parameters con-
			   tained in data and fflags.

			   There is a system  wide  limit  on  the  number  of
			   timers     which	is     controlled    by	   the
			   kern.kq_calloutmax sysctl.
```
[Source](https://man.freebsd.org/cgi/man.cgi?kqueue)

I hope this answers your question.

src/io_notifier.cpp
```diff
@@ -0,0 +1,256 @@
...
+    ts.it_value.tv_sec  = 0;
+    ts.it_value.tv_nsec = 0;
+    return ::timerfd_settime(timer.m_fd, 0, &ts, nullptr) != -1;
+#endif
+}
+
+auto io_notifier::next_events(std::chrono::milliseconds timeout)
+    -> std::vector<std::pair<detail::poll_info*, coro::poll_status>>
+{
+    auto ready_events = std::vector<std::pair<detail::poll_info*, coro::poll_status>>();
```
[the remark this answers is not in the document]
tglane: Yeah you are right. Maybe it would be sufficient to just replace `std::vector` with `std::array` here? With that we would just have a stack "allocation" which basically costs nothing and we would not have to deal with cleaning anything up later. Would that be ok for you?

src/io_scheduler.cpp
```diff
@@ -34,17 +29,15 @@ auto io_scheduler::make_shared(options opts) -> std::shared_ptr<io_scheduler>
         s->m_thread_pool = std::make_unique<thread_pool>(std::move(s->m_opts.pool));
     }
 
-    epoll_event e{};
-    e.events = EPOLLIN;
-
-    e.data.ptr = const_cast<void*>(m_shutdown_ptr);
-    epoll_ctl(s->m_epoll_fd, EPOLL_CTL_ADD, s->m_shutdown_fd, &e);
+    auto shutdown_pipe = std::array<fd_t, 2>{};
```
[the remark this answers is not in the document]
tglane: Yes we definitely have to close them all. Thanks for mentioning this!

src/io_notifier.cpp
jbaldwin: Yeah interesting, thank you for providing the documentation. You're right that there isn't stopping anyone externally from using `1` with `kqueue` timers and possibly causing a conflict. Random might be a good way to go? But also might not resolve it, is there any kind of best practices here for libraries for kqueue? Sorry if these questions seem simple I'm just not familiar with it so trying to wrap my head around how this is possibly solved elsewhere.

jbaldwin: I guess  my point is that epoll solves this by having the kernel dish the 'file descriptors' out uniquely per process so the user is guaranteed not to overlap, does kqueue/bsd not have a similar functionality?

src/io_notifier.cpp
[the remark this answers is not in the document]
jbaldwin: Ok, thats reasonable.

src/io_notifier.cpp
tglane: I had the same questions a while ago so it's not simple ;) Actually it seems the best practice is to just use one single global timer instance (i.e. one timer identifier for the program) as we already do here. We could move the `timer_handle` struct into the `coro::detail` namespace to indicate that it is not for enduser use?

jbaldwin: That seems reasonable, assuming the user cannot directly access the kqueue instance or the timer then it should be safe I think

src/io_notifier.cpp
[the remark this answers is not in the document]
jbaldwin: I think with return value optimization a std::array return would be a good choice, but you'll also need to report to the caller how many you filled out. The nice part of the current solution is that the `for(auto [x,y] : next_events())` is quite nice, I'm not sure that will work with a std::array? But I do think I'd rather remove the allocation since that will certainly hurt performance.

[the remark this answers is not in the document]
tglane: I missunderstood your comment. After re-thinking this I would agree to pass in a `std::vector` from the calling `io_scheduler` to reduce the number of allocations.

src/io_notifier.cpp
tglane: Another idea would be to make the `timer_handle` a singleton? I think that would work for the epoll version, too. And with this it would be impossible to to overlap the file descriptor since there is only one global timer handle. What do you think would be a better way of handling this?

jbaldwin: A singleton doesn't work since an application could have multiple `coro::io_scheduler` instances which would each need their own `timer_handle`.

test/bench.cpp
tglane: I just want to let you know that I have implemented all of your comments but this one because the test will not join the server threads.
I will continue to try to find the issue with the test in the following days. I will also push a commit with all the changes I made so far for another round of review. Thanks so far.

src/io_notifier.cpp
tglane: You are right. I chose another approach where we use a static `std::atomic<std::size_t>` that gets incremented with every `coro::detail::timer_handle` as a "identifier generator". This prevents a costly random number generator and should be fine for the amount of `coro::detail::timer_handle` instances will be created.

tglane:
@jbaldwin I found the issue with the test `benchmark tcp::server echo server inline` (line 544).

In line 645 each server thread does this `co_await s.wait_for_clients;` to wait until all connections are done. In each connection task (line 614) the event is set when all connected clients are done. The problem now is that in your test setup all spawned clients connect to port 8080 and all servers are also listening on port 8080. Now if all clients connect to lets say the first server, only here the `co_await` will resume since on all other server threads, no connection is ever established to set the event. Why this only comes to play on macOS is mysterious though, but I'm pretty sure my analysis of the situation here is correct.

I have found two solutions: 
1. Just use a different port for the client-server connections (8080 to 8090?) and everything works fine.
2. Add a check before the `co_await` statement to only await the event in the case that there are active clients connected. Otherwise just return from the server thread.

I think I would favor option 1, because that would guarantee that we actually test with multiple servers listening.

jbaldwin:
Great find, I think the bug is as follows:

There is a race condition on the server task calling accepted.load, then polling and then incrementing the accepted counter if the socket is real. With the connections limit two or more server tasks could get stuck at accepted = 99 and both try and poll for the final connection, one or more never getting a connection and thus producing a hang in the test.

I think the server tasks needs to increment accepted, poll, if real continue, if not try again. If over the connection limit decrement and move on. The problem is the accepted load and increment are not atomic so multiple tasks can enter that poll area. OSIT.

The macos chips might just be so fast compared to my old Intel chips so this manifests rarely for me if at all 😂

tglane:
> Great find, I think the bug is as follows:
> 
> There is a race condition on the server task calling accepted.load, then polling and then incrementing the accepted counter if the socket is real. With the connections limit two or more server tasks could get stuck at accepted = 99 and both try and poll for the final connection, one or more never getting a connection and thus producing a hang in the test.
> 
> I think the server tasks needs to increment accepted, poll, if real continue, if not try again. If over the connection limit decrement and move on. The problem is the accepted load and increment are not atomic so multiple tasks can enter that poll area. OSIT.
> 
> The macos chips might just be so fast compared to my old Intel chips so this manifests rarely for me if at all 😂

I pushed a potential "fix" for the issue with minimal changes for what I think. You might take a look at it. Now all checks of the pipeline are successful for all tested operating systems.

tglane:
> The when any tests don't seem to want to pass, I have a fix in my condvar PR but will extract it tomorrow and get it merged.
> 
> I should be able to look this over in depth tomorrow as well 💯

That sounds good! Let me know when you got the possible fix for the `when_any` tests merged/PR ready. I will rebase my branch and hopefully all tests will pass then because unfortunately  I am not able to reproduce the failure on my local Mac.

jbaldwin:
@tglane some fun times with github actions over the last few days, I think the CI jobs are reasonable stable now, theres still a ring buffer issue I'm working through but if you can rebase and we get everything passing I think we're close to being able to add this in.

I did look at your bench fix and I'm not sure it fully fixes the issue, it just skips waiting on the event, but the real problem I'm pretty sure is that the `accepted.load()` can allow multiple threads/tasks through before the `poll()` returns to increment. I think it should be incremented immediately and if its great than the allowed amount then stop.

tglane:
> @tglane some fun times with github actions over the last few days, I think the CI jobs are reasonable stable now, theres still a ring buffer issue I'm working through but if you can rebase and we get everything passing I think we're close to being able to add this in.

Oh, that's nice. I made a quick rebase, but it seems there is still something more broken I guess. Will take a closer look at it over the weekend.

> I did look at your bench fix and I'm not sure it fully fixes the issue, it just skips waiting on the event, but the real problem I'm pretty sure is that the `accepted.load()` can allow multiple threads/tasks through before the `poll()` returns to increment. I think it should be incremented immediately and if its great than the allowed amount then stop.

You're probably right... I will take a look at it too to make a proper fix.

jbaldwin:
fwiw the error I've been getting seems to be a github runners issue, any kind of weird `apt-get` incantations with cleaning and re-updating doesn't seem to fix it, it appears to be some kind of github firewall/proxy issue, I just don't know. It was working fine like an hour ago and it just hit your PR again :angry: 

`E: Failed to fetch http://archive.ubuntu.com/ubuntu/pool/main/u/unattended-upgrades/unattended-upgrades_2.8ubuntu1_all.deb  403  Forbidden [IP: 185.125.190.83 80]`

Just re-running ubuntu jobs until you get on a non-broken host seems to fix it :facepalm:

src/io_scheduler.cpp
```diff
@@ -65,21 +58,29 @@ io_scheduler::~io_scheduler()
...
-    if (m_schedule_fd != -1)
+    if (m_schedule_fd[1] != -1)
     {
-        close(m_schedule_fd);
-        m_schedule_fd = -1;
+        close(m_schedule_fd[1]);
+        m_schedule_fd[1] = -1;
     }
+
+    // TODO Close all pipe-fds
```
jbaldwin, line 83: are these all handled?

jbaldwin: Thanks.

I'm looking into some of the failures too, I _think_ they are unrelated, but I'm also having a hard time reproducing locally. I've been running the two fedora failures locally for about an hour and cannot get it to trigger :grimacing:

jbaldwin: > Great! I appreciate your help with that :)
> 
> I figured that I can only trigger the segfaults _without_ sanitizers enabled on my Mac but I will keep looking into that. I really want to get this merged after all the work at some point :D

Totally, its a very nice change. I'm probably going to open bugs for the fedora/conan issues and track those separately is my current though.

jbaldwin: It appears it might be indeed. I can merge that portion of the fix and we can see if this dissapears. There is another problem when a `when_any` test as well with a bad pointer free/segfault I'm still working through, but let me split that off and get the 270 fix re-merged.

tglane: Oh perfect, it seems to fix the segfault and timeout problems for macOS completely! I will rebase the commits into one commit and check if I need to make changes for the static code analysis job to complete and then this should be ready. Thanks a lot!

src/io_scheduler.cpp
```diff
@@ -323,8 +296,8 @@ auto io_scheduler::process_scheduled_execute_inline() -> void
         tasks.swap(m_scheduled_tasks);
 
         // Clear the schedule eventfd if this is a scheduled task.
-        eventfd_t value{0};
-        eventfd_read(m_schedule_fd, &value);
+        int control = 0;
+        ::read(m_schedule_fd[1], reinterpret_cast<void*>(&control), sizeof(control));
```
jbaldwin, line 300: codacy is complaining the return value of `::read` isn't used, but I never checked `eventfd_read` before either, I think this is reasonably safe since we guarantee we've written to it via the `m_schedule_fd_triggered` atomic.

jbaldwin (approved):

src/io_scheduler.cpp
jbaldwin: I'll resolve this so you can merge it, but just wanted to make a note of it.

jbaldwin:
I'll let you do the merge, I usually squash + merge so its a nice atomic commit on `main`.

jbaldwin:
Ah weird, I thought I had it setup so if it's passing with an approval the author can merge, I'll revisit the settings later.

PR #318, opened 2025-05-09
Update mac os ci workflow to use newer version of llvm and mac os

tglane:
This PR should fix the CI for MacOS. I had to add some missing headers and I had to make some changes to the move constructor of `coro::queue` that you should review because I honestly have no idea how you got that to compile prior to my changes (your mutex class is not move constructible but is moved in the move constructor of queue).

### DIFF
--- a/include/coro/queue.hpp
+++ b/include/coro/queue.hpp
@@ -4,7 +4,6 @@
 #include "coro/expected.hpp"
 #include "coro/sync_wait.hpp"
 
-#include <mutex>
 #include <queue>
 
 namespace coro
@@ -30,7 +29,6 @@ enum class queue_consume_result
     queue_stopped
 };
 
-
 /**
  * @brief An unbounded queue. If the queue is empty and there are waiters to consume then
  *        there are no allocations and the coroutine context will simply be passed to the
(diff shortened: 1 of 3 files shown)


include/coro/queue.hpp
```diff
@@ -147,25 +142,17 @@ class queue
...
     auto operator=(queue&& other) -> queue&
     {
         if (std::addressof(other) != this)
         {
-            m_waiters  = std::exchange(other.m_waiters, nullptr);
-            m_mutex    = std::move(other.m_mutex);
-            m_elements = std::move(other.m_elements);
-            m_shutting_down = std::move(other.m_shutting_down);
-            m_stopped = std::move(other.m_stopped);
+            m_waiters       = std::exchange(other.m_waiters, nullptr);
```
jbaldwin, line 152: Unfortunately I think the move constructor and move assignment operator are broken without being able to move the `coro::mutex`, we could put it in a `std::unique_ptr<coro::mutex.` to make it properly move-able at the cost of an extra allocation OR just disable moving and force the user to place it into a `std::unique_ptr` if they need to be able to move it around. I really don't know why this compiled on all platforms previously because it should absolutely not have been allowed and I definitely missed it.

jbaldwin (changes requested):

jbaldwin:
You are absolutely correct.. that should not have compiled before 😬 I left a note on the move constructor/operator.. preference is probably to just `delete` them?

include/coro/queue.hpp
tglane: Yeah, you are right. Just creating a new mutex in the move constructor will break the queue.
I think the main question to answer is whether it is important that the `coro::queue` type is movable?

Since your other collection type - `coro::ring_buffer` - is not movable too, I would tend to deleting the move semantics for `coro::queue` as well.

jbaldwin (approved):

tglane:
Hmm seems like one test timed out without actually running the tests? I think this is not related to the changes made in this PR.

PR #321, opened 2025-05-24
Add coro::default_executor

dobord:
Add a common default executor for use both in the library (e.g. libcoro) and in code that uses libcoro


jbaldwin (approved):
Really like the idea of the perfect() call, nice way to give a sensible default but also let the user choose specifically if they want to.

jbaldwin (approved):

PR #336, opened 2025-06-03
coro::ring_buffer use coro::mutex instead of std::mutex

### LINKED ISSUE #335: ring_buffer use coro::mutex instead of std::mutex
When writing coro::queue it became clear the mutex can be coro::mutex instead of std::mutex, lets make the same change to coro::ring_buffer so its more properly async.

jbaldwin:
* Replace the coro::ring_buffer std::mutex with coro::mutex
* Awaiting coroutines will now properly suspend instead of block the thread if the lock isn't available.

Closes #335


test/test_ring_buffer.cpp
```diff
@@ -65,6 +65,7 @@ TEST_CASE("ring_buffer many elements many producers many consumers", "[ring_buff
 
         for (size_t i = 1; i <= to_produce; ++i)
         {
+            // std::cerr << "co_await rb.produce(" << i << ")\n";
```
jbaldwin, line 68: delete

PR #344, opened 2025-06-10
add internal "std::mutex" that is never contested

jbaldwin:


jbaldwin:
I'm not sure this is required but going to keep it as a draft since its an "idea".

PR #354, opened 2025-06-21
task_container: add back task_container

ripose-jp:
Adds back `coro::task_container`. Since this has been removed, there has been no easy way to share a `thread_pool`/`io_scheduler` across an application, while retaining the ability to await the completion of a subset of tasks.

Task containers are also useful in cases where a class starts a task on a thread pool that references itself. If the task lives longer than the class, that reference begins to dangle. Task containers solve this by allowing the class to `yield_until_empty()` in the destructor to avoid dangling references.

Task containers can also function as latches in cases where `coro::when_all` is insufficient. If the data required to start multiple tasks becomes available at different times, tasks can be started immediately on the task container and then `yield_until_empty()` can function as the latch when necessary.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -95,8 +95,8 @@ set(LIBCORO_SOURCE_FILES
     include/coro/attribute.hpp
     include/coro/condition_variable.hpp src/condition_variable.cpp
     include/coro/coro.hpp
-    include/coro/event.hpp src/event.cpp
     include/coro/default_executor.hpp src/default_executor.cpp
+    include/coro/event.hpp src/event.cpp
     include/coro/generator.hpp
     include/coro/latch.hpp
     include/coro/mutex.hpp src/mutex.cpp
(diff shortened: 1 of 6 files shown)


jbaldwin:
Thank you for the details I think I have a better understanding and just didn't consider these use cases.

It's kinda like a latch but not exactly since it needs to track task lifetimes too, and using the destructor doesn't make sense because this happens many times during the lifetime of the application. Am I understanding a bit better?

Also understand on the thread pool/io scheduler, no problem.

ripose-jp:
>It's kinda like a latch but not exactly since it needs to track task lifetimes too, and using the destructor doesn't make sense because this happens many times during the lifetime of the application. Am I understanding a bit better?

This is exactly right.

I was thinking about it a little bit and if there was some way to have tasks eagerly started on a thread pool while still returning an awaitable, I could accomplish the same thing. I think I like the task container more though since I'm already used to it and doesn't require adding a whole new coroutine type.

jbaldwin:
I'd be curious if this https://github.com/jbaldwin/libcoro/blob/main/include%2Fcoro%2Fthread_pool.hpp#L136 schedule would work with a when_all()?

I added this to thread pool and io scheduler recently when the task container was removed.

ripose-jp:
I mention that function in my second comment. The issue with it is that it's lazy while `spawn` and `start` are eager.

jbaldwin:
Let me think on if it's possible to add a schedule that isn't lazy but returns a task you can await, would that satisfy your requirement?

ripose-jp:
That would make things better. My worry is that I'll need to await a task before I know it's safe to discard. Something like a `done() -> bool` function would be enough to fix this.

jbaldwin (approved):

PR #372, opened 2025-08-14
cmake(android): skip linking -lpthread on Android

dobord:
On Android pthread symbols are provided by bionic libc; there is no separate libpthread. Adding -lpthread triggers a link error (-lpthread not found).

Guard pthread linkage with (UNIX AND NOT ANDROID) and document the reason. Other UNIX platforms (Linux, *BSD) still link against pthread explicitly; no behavior change elsewhere.

### DIFF
--- a/.githooks/readme-template.md
+++ b/.githooks/readme-template.md
@@ -46,6 +46,7 @@
 *
 * [Requirements](#requirements)
 * [Build Instructions](#build-instructions)
+* [Android Support](#android-support)
 * [Contributing](#contributing)
 * [Support](#support)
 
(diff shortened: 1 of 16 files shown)


jbaldwin (approved):
Would you be open to adding a ci-android.yml file to make android and official build target?

jbaldwin:
```
09-03 13:47:04.566 I/coroTest( 3056): BEGIN when_any two tasks one long running with cancellation
09-03 13:47:04.672 I/coroTest( 3056): yielding with amount=1
09-03 13:47:04.749 I/coroTest( 3056): not throwing
09-03 13:56:51.376 I/coroTest( 3056): Global timeout reached, detaching test runner thread...
```

I wonder if this implies there is still a race condition on this test, its been fixed multiple times already, or if there is something different on the platform which requires stronger guarantees on setting the return value

jbaldwin (dismissed):
If you want to add android to the `readme-template.md` as a supported platform that would be appropriate!

jbaldwin:
I think you'll need to rebase onto main or merge main into your branch, there was a ci change which now requires `fetch-depth: 0` and `fetch-tags: true` true on the checkout step for each ci file.

This change is awesome, I had no idea how much it would be to get the tests to actually run, but very nice job!

jbaldwin (approved):

PR #374, opened 2025-08-20
Properly setup socket on linux

mhermier-pharmax:
The setsockopt takes an enum as optname, not a binary flag, contary to what the code does. Fix the socket setup code to call setsockopt for each altered options.


mhermier-pharmax:
That version is slightly different than the original patch. In addition to the setup changes, the comments and exceptions messages were updated to reflect the new code.

jbaldwin:
@mhermier-pharmax thanks for creating the PR!

PR #392, opened 2025-09-12
TCP+TLS Client's use mutable|const buffers for send/recv return values

### LINKED ISSUE #390: tcp client buffer concept only supports 'char', I would like to use 'uint8_t'
In my application I expect `uin8_t` data over the tcp port (basically a byte stream). 

However the current tcp client only supports a `char` based buffer (`std:::string` or `std::vector<char>`).

Could the buffer concept be made less-restrictive, and also support `std::vector<uint8_t>`?

I let AI suggest some code changes:
```cpp
#include <cstdint>
#include <string>
#include <vector>
#include <type_traits>

template <typename type>
concept const_buffer = requires(const type t)
{
    // The underlying data type must be a standard-layout type and trivially copyable
    typename std::remove_pointer_t<decltype(t.data())>;
    requires std::is_trivial_v<std::remove_pointer_t<decltype(t.data())>>;

    // General buffer properties
    { t.empty() } -> std::same_as<bool>;
    { t.size() } -> std::same_as<std::size_t>;

    // We check the return type of `data()` to be a const pointer to the underlying type
    { t.data() } -> std::same_as<const typename std::remove_pointer_t<decltype(t.data())>*>;
};

//----------------------------------------------------------------------------------------------------

template <typename type>
concept mutable_buffer = requires(type t)
{
    // The underlying data type must be a standard-layout type and trivially copyable
    typename std::remove_pointer_t<decltype(t.data())>;
    requires std::is_trivial_v<std::remove_pointer_t<decltype(t.data())>>;

    // General buffer properties
    { t.empty() } -> std::same_as<bool>;
    { t.size() } -> std::same_as<std::size_t>;

    // We check the return type of `data()` to be a non-const pointer to the underlying type
    { t.data() } -> std::same_as<typename std::remove_pointer_t<decltype(t.data())>*>;
};

//----------------------------------------------------------------------------------------------------

// Example Usage & Demonstration

struct CharBuffer {
    bool empty() const { return false; }
    const char* data() const { return nullptr; }
    std::size_t size() const { return 0; }
    char* data() { return nullptr; }
};

struct Uint8Buffer {
    bool empty() const { return false; }
    const uint8_t* data() const { return nullptr; }
    std::size_t size() const { return 0; }
    uint8_t* data() { return nullptr; }
};

void check_concepts() {
    static_assert(const_buffer<CharBuffer>);
    static_assert(mutable_buffer<CharBuffer>);

    static_assert(const_buffer<Uint8Buffer>);
    static_assert(mutable_buffer<Uint8Buffer>);
    
    // std::vector should now also satisfy the concepts
    static_assert(const_buffer<std::vector<char>>);
    static_assert(mutable_buffer<std::vector<char>>);

    static_assert(const_buffer<std::vector<uint8_t>>);
    static_assert(mutable_buffer<std::vector<uint8_t>>);
}
```

The new AI generated implementation of recv:
```cpp
template<concepts::mutable_buffer buffer_type>
auto recv(buffer_type&& buffer) -> std::pair<recv_status, std::span<typename std::remove_pointer_t<decltype(buffer.data())>>>
{
    // If the user requested zero bytes, just return.
    if (buffer.empty())
    {
        using element_type = std::remove_pointer_t<decltype(buffer.data())>;
        return {recv_status::ok, std::span<element_type>{}};
    }

    // Cast the data pointer to a void* for the C-style API,
    // which is the correct and safe way to handle a generic pointer.
    auto bytes_recv = ::recv(m_socket.native_handle(), static_cast<void*>(buffer.data()), buffer.size(), 0);

    if (bytes_recv > 0)
    {
        // Ok, we've received some data.
        using element_type = std::remove_pointer_t<decltype(buffer.data())>;
        return {recv_status::ok, std::span<element_type>{buffer.data(), static_cast<size_t>(bytes_recv)}};
    }
    else if (bytes_recv == 0)
    {
        // On TCP stream sockets 0 indicates the connection has been closed by the peer.
        return {recv_status::closed, std::span<char>{}};
    }
    else
    {
        // Report the error to the user.
        return {static_cast<recv_status>(errno), std::span<char>{}};
    }
}
```

jbaldwin:
* This adds another templated parameter to send/recv for tls|tcp clients on the return value type so the returned buffers contain the same element_type instead of just char.

Closes #390

### DIFF
--- a/include/coro/concepts/buffer.hpp
+++ b/include/coro/concepts/buffer.hpp
@@ -22,6 +22,12 @@ concept const_buffer = requires(const type t)
     { t.data() } -> std::same_as<const typename std::remove_pointer_t<decltype(t.data())>*>;
 };
 
+template<const_buffer buffer_type>
+struct const_buffer_traits
+{
+    using element_type = std::add_const_t<std::remove_pointer_t<decltype(std::declval<buffer_type>().data())>>;
+};
+
 template<typename type>
 concept mutable_buffer = requires(type t)
 {
@@ -36,6 +42,13 @@ concept mutable_buffer = requires(type t)
     // We check the return type of `data()` to be a non-const pointer to the underlying type
     { t.data() } -> std::same_as<typename std::remove_pointer_t<decltype(t.data())>*>;
 };
+
+template<mutable_buffer buffer_type>
+struct mutable_buffer_traits
+{
+    using element_type = std::remove_pointer_t<decltype(std::declval<buffer_type>().data())>;
+};
+
 // clang-format on
 
 } // namespace coro::concepts
--- a/include/coro/net/tcp/client.hpp
+++ b/include/coro/net/tcp/client.hpp
@@ -30,7 +30,7 @@ public:
     };
 
     /**
-     * Creates a new tcp client that can connect to an ip address + port.  By default the socket
+     * Creates a new tcp client that can connect to an ip address + port. By default, the socket
      * created will be in non-blocking mode, meaning that any sending or receiving of data should
      * poll for event readiness prior.
      * @param scheduler The io scheduler to drive the tcp client.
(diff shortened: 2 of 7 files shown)


borro0:
That looks good, perhaps add a test with std::vector<uint8_t> as buffer, to see if it works? You could consider to include udp `peer.hpp` as well.

PR #397, opened 2025-10-01
Avoid reference cycles in thread_pool and io_schedule

### LINKED ISSUE #396: Thread pools are never destroyed due to reference cycle
If I do:

```
auto exec = coro::thread_pool::make_shared(...);
```

to obtain a shared pointer to a thread pool, even when my reference goes out of scope, the dtor for the `thread_pool` is never run.

This appears to be because the `std::vector` of threads that the thread pool each run with a function that hold a strong reference to the same `exec` shared pointer:

```
auto thread_pool::make_shared(options opts) -> std::shared_ptr<thread_pool>
{
    auto tp = std::make_shared<thread_pool>(std::move(opts), private_constructor{});

    // Initialize once the shared pointer is constructor so it can be captured for
    // the background threads.
    for (uint32_t i = 0; i < tp->m_opts.thread_count; ++i)
    {
        tp->m_threads.emplace_back([tp, i]() { tp->executor(i); });
    }

    return tp;
}
```

Since `m_threads` is a member of `tp` we have a reference cycle `tp -> m_threads -> lambda -> tp`.

Consequently, the refcount of this shared pointer can never drop to zero, unless we explicitly call `thread_pool::shutdown()`. Is this the intended usage?

Is there a particular reason that the background threads need to keep the thread pool alive? Or should it rather be that the thread pool must be kept alive by the user for the lifetime of its use?

If the latter then perhaps one just needs to do?
```diff
diff --git a/src/thread_pool.cpp b/src/thread_pool.cpp
index 6740894..a7ba142 100644
--- a/src/thread_pool.cpp
+++ b/src/thread_pool.cpp
@@ -26,7 +26,7 @@ auto thread_pool::make_shared(options opts) -> std::shared_ptr<thread_pool>
     // the background threads.
     for (uint32_t i = 0; i < tp->m_opts.thread_count; ++i)
     {
-        tp->m_threads.emplace_back([tp, i]() { tp->executor(i); });
+        tp->m_threads.emplace_back([tp = tp.get(), i]() { tp->executor(i); });
     }
 
     return tp;
```

wence-:
Previously, the background threads in a thread_pool and/or io_scheduler ran with a lambda capturing a shared pointer to the relevant instance. This caused a reference cycle such that the refcount of the shared pointer handed back to the user would never go to zero unless they explicitly called shutdown().

To fix this, just capture a raw pointer, this is safe because the dtor for the pool object joins the background threads and hence they can never see a dangling pointer.

Fixes #396.

### DIFF
--- a/src/io_scheduler.cpp
+++ b/src/io_scheduler.cpp
@@ -42,7 +42,8 @@ auto io_scheduler::make_shared(options opts) -> std::shared_ptr<io_scheduler>
     // the background thread.
     if (s->m_opts.thread_strategy == thread_strategy_t::spawn)
     {
-        s->m_io_thread = std::thread([s]() { s->process_events_dedicated_thread(); });
+        // IO thread captures a raw pointer to avoid reference cycles
+        s->m_io_thread = std::thread([s = s.get()]() { s->process_events_dedicated_thread(); });
     }
     // else manual mode, the user must call process_events.
 
--- a/src/thread_pool.cpp
+++ b/src/thread_pool.cpp
@@ -26,7 +26,8 @@ auto thread_pool::make_shared(options opts) -> std::shared_ptr<thread_pool>
     // the background threads.
     for (uint32_t i = 0; i < tp->m_opts.thread_count; ++i)
     {
-        tp->m_threads.emplace_back([tp, i]() { tp->executor(i); });
+        // Threads capture a raw pointer to avoid reference cycles
+        tp->m_threads.emplace_back([tp = tp.get(), i]() { tp->executor(i); });
     }
 
     return tp;
--- a/test/test_io_scheduler.cpp
+++ b/test/test_io_scheduler.cpp
@@ -890,7 +890,7 @@ TEST_CASE("io_scheduler destruction with io thread", "[io_scheduler]")
         weakref = scheduler;
         REQUIRE(weakref.lock() != nullptr);
     }
-    REQUIRE_FALSE(weakref.lock() == nullptr);
+    REQUIRE(weakref.lock() == nullptr);
 }
 
 TEST_CASE("~io_scheduler", "[io_scheduler]")
--- a/test/test_io_scheduler.cpp
+++ b/test/test_io_scheduler.cpp
@@ -5,8 +5,9 @@
 
 #include <atomic>
 #include <chrono>
-#include <thread>
 #include <iostream>
+#include <memory>
+#include <thread>
 
 #include <cstring>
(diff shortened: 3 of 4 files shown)


wence-:
Hmm, the tcp_server graceful shutdown via socket test fails in many of the CI runs, let me try and repro and take a look tomorrow, unless you've got any good ideas.

jbaldwin:
I'm away this weekend but I'll give it a good look early next week.

jbaldwin:
Running this through valgrind now, I haven't done that in a while since the tests take a very long time.

jbaldwin:
```
===============================================================================
All tests passed (2104381 assertions in 219 test cases)

==9298== 
==9298== HEAP SUMMARY:
==9298==     in use at exit: 0 bytes in 0 blocks
==9298==   total heap usage: 151,870,367 allocs, 151,870,367 frees, 17,025,454,324 bytes allocated
==9298== 
==9298== All heap blocks were freed -- no leaks are possible
==9298== 
==9298== For lists of detected and suppressed errors, rerun with: -s
==9298== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```

wence-:
> I think my only question right now is if the `std::weak_ptr` change across all the objects that use an executor is needed? I would think if an application spins down resources correctly in the right order it shouldn't be necessary to have the `std::weak_ptr` there. But I could certainly see it being easy to destroy the executor `std::shared_ptr` and let all the objects throw `std::bad_weak_ptr`. I'm just worried about the perf cost of checking the pointer at these junctions a lot if it isn't necessary.

I don't think it's necessary to use a weak pointer that one upgrades when needed. I wanted that so I could "easily" detect a dead reference rather than passing a raw pointer and getting use-after-free if there were an error.

I suppose that one could instead pass a `std::shared_ptr<executor>&` reference so if the executor shared pointer is destroyed we can detect things but we don't increment the refcount.

jbaldwin:
```
1: terminate called after throwing an instance of 'std::system_error'
1:   what():  Resource deadlock avoided
```

Well ok, this is interesting and I feel a bit dumb in retrospect because this is pretty obvious. The executor's destructor is attempting to join our thread on the thread that is running it, likely due to the last std::shared_ptr<> ref being within one of the coroutines.

I used to have these all as std::unique_ptr<coro::thread_pool|coro::io_scheduler> and converted them to std::shared_ptr<> for safety around all the objects using it/referencing the executors (thread_pool/scheduler) and it seems like this is the consequence and the circular ref was hiding it.

Issue #401, opened 2025-10-16
`ring_buffer::shutdown_drain()` deadlocks when channel is forcibly shut down


pentschev:
# Summary

The `ring_buffer::shutdown_drain()` method can deadlock when a channel is forcibly shut down (via `shutdown()`) while a drain operation is in progress. This occurs when one coroutine is draining the buffer while another coroutine dies unexpectedly and triggers a shutdown, leaving the draining coroutine stuck in an infinite loop.

# Root Cause

`shutdown_drain` only checks if the buffer is empty:

https://github.com/jbaldwin/libcoro/blob/7e0ce982405fb26b6ca8af97f40a8eaa2b78c4fa/include/coro/ring_buffer.hpp#L304-L307

This loop assumes that consumers will eventually drain all messages. However, if `shutdown()` is called by another coroutine (changing `m_running_state` from `draining` to `stopped`), the draining coroutine never detects this state change and continues waiting indefinitely for the buffer to become empty, which will never happen if all consumers are dead.

# Minimal Reproducer

Here's an example demonstrating the deadlock:

```cpp
#include <iostream>
#include <memory>
#include <stdexcept>

#include <coro/coro.hpp>

coro::task<void> producer(std::shared_ptr<coro::ring_buffer<int, 1>> rb,
                          std::unique_ptr<coro::thread_pool>& executor) {
    co_await executor->schedule();

    co_await rb->produce(42);
    std::cout << "Producer: calling shutdown_drain\n";
    co_await rb->shutdown_drain(executor);  // DEADLOCKS HERE
    std::cout << "Producer: drain complete\n";
}

coro::task<void> consumer(std::shared_ptr<coro::ring_buffer<int, 1>> rb,
                          std::unique_ptr<coro::thread_pool>& executor) {
    co_await executor->schedule();
    std::cout << "Consumer: forcefully shutting down\n";
    co_await rb->shutdown();  // This, not a throw
    std::cout << "Consumer: shutdown complete\n";
}

int main() {
    auto executor = coro::thread_pool::make_unique(
        coro::thread_pool::options{.thread_count = 1});
    auto rb = std::make_shared<coro::ring_buffer<int, 1>>();

    auto result = coro::sync_wait(coro::when_all(
        producer(rb, executor),
        consumer(rb, executor)
    ));

    return 0;
}
```

# Proposed Solution

Add a check for the shutdown state in the drain loop:

```diff
--- a/include/coro/ring_buffer.hpp
[...]

PR #402, opened 2025-10-17
Fix deadlock in ring_buffer and queue shutdown_drain

wence-:
If a ring_buffer (queue) produces an element such that the buffer is
non-empty and then calls shutdown_drain we enter a draining loop where we
yield until the ring buffer is empty. If, meanwhile, the consumer side of
the buffer decides to forcibly shutdown without consuming the produced
value (for example because it has received an exception that should be
propagated and therefore wants to wake and shutdown the producer) then we
never exit the yield-until-empty loop.

To fix this only loop while the buffer is both non-empty and we are still
in the draining state.

The produced value(s) that are stored in the buffer are now never
processed which I think is the right semantics since the consumer has
explicitly said "I can't do anything with this".

- Fixes https://github.com/jbaldwin/libcoro/issues/401

### DIFF
--- a/test/test_queue.cpp
+++ b/test/test_queue.cpp
@@ -242,10 +242,9 @@ TEST_CASE("queue shutdown_drain non-empty consumer shutdown", "[queue]")
     std::cerr << "BEGIN queue issue-401\n";
 
     coro::queue<int> q{};
-    auto             exec = coro::io_scheduler::make_unique(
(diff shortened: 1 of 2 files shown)


include/coro/queue.hpp
wence-: Drive-by changes because clangd kept on adding these includes automatically in the test file above <coro/coro.hpp> and I was getting build errors.

PR #403, opened 2025-10-17
Store raw pointer in udp::peer and implement copy/move-assignment

wence-:
Previously the peer struct held a reference member that could not be reseated and so the defaulted copy/move-assignments were implicitly deleted. We want these objects to be moveable and copyable so instead store a raw pointer.

### DIFF
--- a/test/test_task_container.cpp
+++ b/test/test_task_container.cpp
@@ -14,19 +14,22 @@ using namespace std::chrono_literals;
 
 TEST_CASE("task_container schedule single task", "[task_container]")
 {
-    auto s = coro::thread_pool::make_unique(
-        coro::thread_pool::options{.thread_count = 1});
-
-    int value = 0;
-    auto make_task = [] (int &value) -> coro::task<void>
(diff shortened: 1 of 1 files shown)


include/coro/net/udp/peer.hpp
```diff
@@ -45,8 +44,8 @@ class peer
 
     peer(const peer&)                             = default;
     peer(peer&&)                                  = default;
-    auto operator=(const peer&) noexcept -> peer& = default;
-    auto operator=(peer&&) noexcept -> peer&      = default;
+    auto operator=(const peer&) noexcept -> peer& = delete;
```
jbaldwin: I think to match `tcp::client` and `tls::client` the scheduler pointer class field should likely be a raw pointer so the `udp::peer` can be properly copied. It appears I missed converting it in the prior PR that switched from `std::shared_ptr` to `std::unique_ptr`, my bad.

Would you mind switching the PR to match the other types in behavior?

jbaldwin (changes requested):

include/coro/net/udp/peer.hpp
wence-: Done.

PR #416, opened 2025-11-24
condition_variable: mutual exclusion for notify/wait

### LINKED ISSUE #398: deadlock/missing wake in condition_variable notify_all?
The code below mimics a hang that we're seeing in a real application where it appears that when I have multiple awaiters suspended waiting on a condition_variable and then run notify_all, one waiter is lost and never woken.

The general flow is that I have some object that is not coroutine-aware but has a callback interface, so it gets a signal, puts some data in a "ready" set and then notifies a coro condition_variable. Meanwhile I have some number of consumer tasks waiting on that condition_variable. When they are woken they extract the data from the ready set and move it to a done set (all with a lock held). The cv predicate is checking whether the ready set has any entries or if the done set is full.

If I compile with `-O0` and then run `while taskset -c 0 ./callback; do /bin/true; done` I usually get a hang in a few iterations.

On the main thread I can see that the `Notifier`s condition variable when hung has a live awaiter:

```
(gdb) p *n.cv.m_awaiters._M_b._M_p
$3 = {_vptr$awaiter_base = 0x55555558aac0 <vtable for coro::condition_variable::awaiter_with_predicate+16>, m_next = 0x0, 
  m_awaiting_coroutine = {_M_fr_ptr = 0x7ffff0000e70}, m_condition_variable = @0x7fffffffd8f8, m_lock = @0x7ffff0000f30}
```

But the thread pool does not have any coroutines in its queue
```
(gdb) p tp->m_queue
$7 = std::deque with 0 elements
```

Are we using condition_variables incorrectly here? Thanks!

<details>
<summary> callback.cxx </summary>

```c++
#include <chrono>
#include <coro/coro.hpp>
#include <iostream>
#include <thread>
#include <unordered_set>

static constexpr int ncallbacks = 1000;
static constexpr int nconsumers = 16;
struct Notifier {
  Notifier(std::shared_ptr<coro::thread_pool> executor) : signals{executor} {
    std::cout << std::this_thread::get_id() << " Notifier()" << std::endl;
  }
  ~Notifier() {
    std::cout << std::this_thread::get_id() << " ~Notifier()" << std::endl;
  }
  coro::task<void> wait_for_signal([[maybe_unused]] std::shared_ptr<coro::thread_pool> executor) {
    std::cout << std::this_thread::get_id() << " wait_for_signal()..."
              << std::endl;
    while (true) {
      auto lck = co_await mtx.scoped_lock();
      co_await cv.wait(
          lck, [this]() { return ready.size() != 0 || done.size() == ncallbacks; });
      std::cout << std::this_thread::get_id() << " wait_for_signal()...woken" << std::endl;
      if (!ready.empty()) {
        auto v = ready.extract(ready.begin());
        done.insert(v.value());
      } else {
        break;
      }
    }
    std::cout << std::this_thread::get_id() << " wait_for_signal()...done"
              << std::endl;
  }
  coro::task<void> signal(int i) {
    {
      auto lock = co_await mtx.scoped_lock();
      ready.insert(i);
      std::cout << std::this_thread::get_id() << " signal()..set flag"
                << std::endl;
    }
    std::cout << std::this_thread::get_id() << " signal()...notifying"
              << std::endl;
    co_await cv.notify_all();
    std::cout << std::this_thread::get_id() << " signal()...notifying done."
              << std::endl;
  }
  void signal_task_container(std::shared_ptr<coro::thread_pool> executor) {
    auto t = std::thread([this, executor]() {
      std::this_thread::sleep_for(std::chrono::milliseconds{100});
      for (int i = 0; i < ncallbacks; i++) {
        signals.start(signal(i));
      }
    });
    t.detach();
  }
  coro::mutex mtx;
  coro::condition_variable cv;
  bool flag{false};
  coro::task_container<coro::thread_pool> signals;
  std::unordered_set<int> ready;
  std::unordered_set<int> done;
};

coro::task<void>
notification_task_container(Notifier& n, std::shared_ptr<coro::thread_pool> executor) {
  std::cout << std::this_thread::get_id() << " notification()..." << std::endl;
  co_await executor->schedule();
  std::cout << std::this_thread::get_id() << " notification()...scheduled"
            << std::endl;
  std::vector<coro::task<void>> consumers;
  for (int i = 0; i < nconsumers; i++) {
    consumers.push_back(executor->schedule(n.wait_for_signal(executor)));
  }
  n.signal_task_container(executor);
  auto results = co_await coro::when_all(std::move(consumers));
  for (auto &r : results) {
    r.return_value();
  }
  std::cout <<  std::this_thread::get_id() << " Made it to yield" << std::endl;
  co_await n.signals.yield_until_empty();
  std::cout << std::this_thread::get_id()
            << " notification()...waited for signal" << std::endl;
}

int main() {
  std::cout << std::this_thread::get_id() << " main()..." << std::endl;
  auto tp = coro::thread_pool::make_shared(
      coro::thread_pool::options{.thread_count = 2});
  auto n = Notifier(tp);
  std::cout << "Notification with task container" << std::endl;
  coro::sync_wait(notification_task_container(n, tp));
  std::cout << std::this_thread::get_id() << " main()...done" << std::endl;
}
```

</details>

[CMakeLists.txt](https://github.com/user-attachments/files/22728485/CMakeLists.txt)

wence-:
To avoid lost wakeups, use an internal mutex to protect modification of the awaiter list.

Now, while we are notifying, if another notification task arrives it must wait for our modification of the awaiter list to complete. Hence waiters that were not ready and are pushed back on the list have a chance to be woken up by the next notification.

- Closes #398

I came back to the issue raised in #398 because we have a use case now where condvars make the most sense and we legitimately might have multiple simultaneous notification calls.

Since the modification awaiter list representing the suspended awaiters waiting for a notification crosses coroutine boundaries, I don't think we can protect it with a `std::mutex`. Hence I use a `coro::mutex` (internally managed by the condvar).

### DIFF
--- a/include/coro/condition_variable.hpp
+++ b/include/coro/condition_variable.hpp
@@ -6,7 +6,7 @@
 #include "coro/task.hpp"
 #include "coro/event.hpp"
 #include "coro/mutex.hpp"
-#include "coro/when_any.hpp"
+#include "coro/when_all.hpp"
 
 #include <atomic>
 #include <chrono>
@@ -39,7 +39,7 @@ private:
 
     struct awaiter_base
     {
-        awaiter_base(coro::condition_variable& cv, coro::scoped_lock& l);
+        awaiter_base(coro::condition_variable& cv, coro::scoped_lock& l, coro::scoped_lock notify_lock);
         virtual ~awaiter_base() = default;
 
         awaiter_base(const awaiter_base&) = delete;
@@ -55,6 +55,8 @@ private:
         coro::condition_variable& m_condition_variable;
         /// @brief The lock that the wait() was called with.
         coro::scoped_lock& m_lock;
+        /// @brief Lock used for mutual exclusion between notifiers and waiters
+        coro::scoped_lock m_notify_lock;
 
         /// @brief Each awaiter type defines its own notify behavior.
         /// @return The status of if the waiter's notify result.
@@ -63,7 +65,7 @@ private:
 
     struct awaiter : public awaiter_base
     {
-        awaiter(coro::condition_variable& cv, coro::scoped_lock& l) noexcept;
+        awaiter(coro::condition_variable& cv, coro::scoped_lock& l, coro::scoped_lock notify_lock) noexcept;
         ~awaiter() override = default;
 
         awaiter(const awaiter&) = delete;
@@ -83,6 +85,7 @@ private:
         awaiter_with_predicate(
             coro::condition_variable& cv,
             coro::scoped_lock& l,
+            coro::scoped_lock notify_lock,
             predicate_type p
         ) noexcept;
         ~awaiter_with_predicate() override = default;
@@ -109,6 +112,7 @@ private:
         awaiter_with_predicate_stop_token(
             coro::condition_variable& cv,
             coro::scoped_lock& l,
+            coro::scoped_lock notify_lock,
             predicate_type p,
             std::stop_token stop_token
         ) noexcept;
@@ -183,6 +187,7 @@ private:
         awaiter_with_wait_hook(
             coro::condition_variable& cv,
             coro::scoped_lock& l,
+            coro::scoped_lock notify_lock,
             controller_data& data
         ) noexcept;
         ~awaiter_with_wait_hook() override = default;
@@ -199,11 +204,12 @@ private:
             std::unique_ptr<io_executor_type>& executor,
             coro::condition_variable& cv,
             coro::scoped_lock& l,
+            coro::scoped_lock notify_lock,
             const std::chrono::nanoseconds wait_for,
             std::optional<predicate_type> predicate = std::nullopt,
             std::optional<std::stop_token> stop_token = std::nullopt
         ) noexcept
-            : awaiter_base(cv, l),
+            : awaiter_base(cv, l, std::move(notify_lock)),
               m_executor(executor),
               m_wait_for(wait_for),
               m_predicate(std::move(predicate)),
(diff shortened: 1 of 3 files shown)


jbaldwin:
@wence- I've played around with renaming to `task_group` and make it more ergonomic in that you have to give it the full set of tasks upfront. There were definitely some race conditions on the `yield_until_empty()` if you keep adding tasks into it. I think giving it the full set of tasks upfront makes sense? This way its basically a dynamic lifetime task tracker backed by a coro::latch so the `co_await task_group.wait()` method is super efficient now.

Let me know your thoughts and if this works for the use case you are considering (I think it should for the cond var change).

task_group PR
https://github.com/jbaldwin/libcoro/pull/418

jbaldwin:
https://github.com/jbaldwin/libcoro/pull/419

Heres a stab at getting it to work with the upgraded `task_group`, I think this should solve the wait()/notify() race conditions.

wence-:
One thing I would note with the `task_group` change, I was using a `task_container` morally like this:

```
struct ThingThatReceivesCallbacks {
  coro::task<void> async_notify(coro::latch& latch, args) {
      do_stuff_with(args);
      latch.count_down();
  }

  // This is called from a background thread
  void bridge_from_sync_code(...) {
     task_container_.start(async_notify(latch_, ...));
  }

  coro::task<void> wait_for_all_notifications(...) {
     co_await latch_;
     // This is so the coroutine unwinding doesn't jump into our dtor
     // before the final `async_notify` task completes.
     co_await task_container_.yield_until_empty();
  }
};
```

I think with the new `task_group` I can't do this, because I don't have the tasks up front, I only know how many there are.

I think I can instead do:

```
struct ThingThatReceivesCallbacks {
  coro::task<void> async_notify(coro::latch& latch, args) {
      do_stuff_with(args);
      latch.count_down();
  }

  // This is called from a background thread
  void bridge_from_sync_code(std::unique_ptr<thread_pool>& executor) {
     executor.spawn(async_notify(latch_, ...));
  }

  coro::task<void> wait_for_all_notifications(std::unique_ptr<thread_pool>& executor, ...) {
     co_await latch_;
     // This is so the coroutine unwinding doesn't jump into our dtor
     // before the final `async_notify` task completes.
     co_await executor.yield();
  }
};
```

I think that's equivalent because I can be guaranteed that I only need to yield once to put myself at the back of the task queue to ensure that the final `async_notify` completes before I go out of scope.

jbaldwin:
I re-added the task_group::start(task) method and managed to figure out how to do the tests without the race condition, so if you want to re-update this PR with the updated task_group we can move it forward again.

Sorry for the turbulence in the api, I think we've landed on something good now though.

PR #418, opened 2025-11-25
rename task_container -> task_group

### LINKED ISSUE #417: rename task_container to task_group
container doesn't seem like its the right word, group seems a little more obvious what its for, grouping a set of tasks together.

jbaldwin:
Groups now require the tasks they will track to be given in the constructor, this means they are effectively a dynamic lifetime latch.

Closes #417


include/coro/concepts/executor.hpp
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "coro/concepts/awaitable.hpp"
+#include "coro/concepts/range_of.hpp"
```
jbaldwin: delete

Tearnote:
I'm affected by this breaking change - I currently have a class that schedules all of its fire-and-forget tasks at arbitrary times into a long-lived `coro::task_container`, so that it can then use `.empty()` to trivially communicate if any background work is currently running. Since with `coro::task_group` all tasks now have to be provided at construction time, it's unsuitable for this purpose.

Is there a way to implement the same functionality with any of the other libcoro features? It would be preferable if I didn't have to manually bump an atomic each time I start/complete a task, and I also can't see a way to have a task automatically delete itself from some container when finished.

And, lastly - thank you for libcoro!

jbaldwin:
@Tearnote I've got a PR up #425 if you want to take a look at it, I think I've found a way to get around the race condition on the test and re-introduce the start(task) method, it just might require the user to be careful about `co_await group;` because depending on when thats done it could drop to 0 before all the task start, I've made a note of that on the classes comments.

PR #419, opened 2025-11-26
Jbaldwin/condvar lost wake task group

jbaldwin:


include/coro/condition_variable.hpp
```diff
@@ -394,32 +385,29 @@ class condition_variable
...
-            awaiter_base* next = waiter->m_next;
-            // This will kick off each task in parallel on the scheduler, they will fight over the lock
-            // but will give the best parallelism scheduling them immediately.
-            executor->spawn(make_notify_all_executor_individual_task(waiter));
-            waiter = next;
+            notify_tasks.emplace_back(make_notify_all_executor_individual_task(waiter));
+            waiter = waiter->m_next;
         }
 
+        coro::task_group group{executor, std::move(notify_tasks)};
```
jbaldwin: I think this could possibly also just be a `when_all(notify_tasks)` but we'd need to refactor how `make_notify_all_executor_individual_task()` work internally to possibly take an executor and if it receives one to schedule itself.. which honestly is probably more efficient since it won't introduce the `task_self_deleting` as an intermediate, just a schedule co_await.

jbaldwin: updated with another patchset to show when_all usage, I think they are _roughly_ equivalent, but when_all seems a little more appropriate since we actually do know the lifetime of the tasks, and task_group should be use for more dynamic lifetimes

PR #420, opened 2025-11-28
`thread_pool::shutdown()`: avoid self join

madsbk:
Avoid deadlock or `Resource deadlock avoided` if shutdown is called from a thread in the thread_pool.


madsbk:
I don’t think this is sufficient for my use case, I need something more like: https://github.com/jbaldwin/libcoro/issues/421

PR #423, opened 2025-12-01
executor->spawn(joinable)->task

### LINKED ISSUE #421: [Feature Request] Support spawning non-detached tasks
I have a use case where I need to schedule a coroutine task on a `coro::thread_pool` *immediately*, without awaiting it at submission time. Later during shutdown, I need to `co_await` that same task to ensure it has fully completed before resources are torn down.

Is this already possible? `thread_pool::spawn()` always creates a detached task, which cannot be awaited later. This makes it difficult to safely synchronize with task completion during shutdown and can lead to resource-lifetime issues.

What I need is:

* Submit a task to the thread pool without awaiting it immediately, and
* Retain an awaitable handle that can be awaited at a later point.

Essentially, this would behave like `spawn()` but without detaching, returning a `coro::task<T>` (or similar awaitable) representing the scheduled work. This would allow background tasks to be scheduled safely while still guaranteeing deterministic shutdown.

jbaldwin:
* Rename spawn() to spawn_detached(task)
* Add spawn_joinable(task) -> task

The user can now spawn a joinable task that doesn't have to be co_await'ed inline like schedule(task) -> task requires.

Closes #421

### DIFF
--- a/include/coro/thread_pool.hpp
+++ b/include/coro/thread_pool.hpp
@@ -122,6 +122,7 @@ public:
 
     /**
      * Spawns the given task to be run on this thread pool, the task is detached from the user and cannot be joined.
+     * @note This method is preferable to `spawn_joinable()` when possible as it has less overhead.
      * @param task The task to spawn onto the thread pool.
      * @return True if the task has been spawned onto this thread pool.
      */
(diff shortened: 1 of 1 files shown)


include/coro/thread_pool.hpp
```diff
@@ -119,11 +121,22 @@ class thread_pool
...
      */
-    auto spawn(coro::task<void>&& task) noexcept -> bool;
+    auto spawn_detached(coro::task<void>&& task) noexcept -> bool;
+
+    /**
+     * Spawns the given task to be run on this thread pool, the task returned must be joined in the future.
+     * @note The returned task shouldn't be co_await'ed immediately, the spawned task is started on the thread
+     *       pool automatically but the returned join task *must* be co_await'ed at some point in the future.
+     *       If you drop the returned task it will hang the thread until the spawned task completes so it is
+     *       highly advisable to co_await the returned join task appropriately.
```
madsbk: Add that `spawn_joinable` is not itself a coroutine. And maybe also that `spawn_detached` is preferable because of a lower overhead, if you just need fire and forget.

madsbk (approved):
Look good to me, just what I need :)

PR #439, opened 2026-01-30
"One `socket_address` to rule them all" & UB fix

PyXiion:
I wanted to test IPv6 support, but found out that it doesn't work and the code is full of UB. Also I introduce a much simpler backend API for `sockaddr`.

**Before (with UB)**
```cpp
sockaddr_in         client{};
constexpr const int len = sizeof(struct sockaddr_in);
net::socket         s{::accept(
    m_accept_socket.native_handle(),
    reinterpret_cast<struct sockaddr*>(&client),
    const_cast<socklen_t*>(reinterpret_cast<const socklen_t*>(&len)))}; // UB here, accept() can change `len` depending on the proto (different for ipv4 & ipv6)
```

**After**
```cpp
auto client_endpoint = net::endpoint::make_uninitialised();
auto [addr, len]     = client_endpoint.native_mutable_data();

net::socket s{::accept(m_accept_socket.native_handle(), addr, len)};
```

The same for the clients, it's much simpler
```cpp
auto [addr, len] = m_endpoint.data();
auto cret = ::connect(m_socket.native_handle(), addr, len);
```

ALso it's more extensible in the future, for example, to implement unix domain sockets we'll need to change only the endpoint class.

I need some feedback on this API =D

Also I suggest to modify all other tests using Catch2 generators, so they can check both ipv4 & ipv6 (or maybe just one ipv6 test will do)
```cpp
const auto endpoint = GENERATE(
    as<coro::net::endpoint>{},
    coro::net::endpoint{"127.0.0.1", 8080},
    coro::net::endpoint{"::1", 8080, coro::net::domain_t::ipv6});
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1261,9 +1261,8 @@ int main()
     {
         // Start by creating a tcp server, we'll do this before putting it into the scheduler so
         // it is immediately available for the client to connect since this will create a socket,
-        // bind the socket and start listening on that socket.  See tcp::server for more details on
-        // how to specify the local address and port to bind to as well as enabling SSL/TLS.
-        coro::net::tcp::server server{scheduler};
+        // bind the socket and start listening on that socket.
+        coro::net::tcp::server server{scheduler, {"127.0.0.1", 8080}};
 
         // Now schedule this task onto the scheduler.
         co_await scheduler->schedule();
@@ -1350,9 +1349,8 @@ int main()
         // Immediately schedule onto the scheduler.
         co_await scheduler->schedule();
 
-        // Create the tcp::client with the default settings, see tcp::client for how to set the
-        // ip address, port, and optionally enabling SSL/TLS.
-        coro::net::tcp::client client{scheduler};
+        // Create the tcp::client
+        coro::net::tcp::client client{scheduler, {"127.0.0.1", 8888}};
 
         // Ommitting error checking code for the client, each step should check the status and
         // verify the number of bytes sent or received.


include/coro/net/endpoint.hpp
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    /**
+     * @brief Extracts the address family from the endpoint.
+     * @return An domain_t object
+     * @throws std::runtime_error If the address family is not supported
+     */
+    [[nodiscard]] auto domain() const -> domain_t
+    {
+        if (m_storage.ss_family == AF_INET)
```
jbaldwin: can you use { } around each if since its the style for this project, thanks

include/coro/net/endpoint.hpp
```diff
@@ -0,0 +1,137 @@
...
+    {
+        return m_len == other.m_len && std::memcmp(&m_storage, &other.m_storage, m_len) == 0;
+    }
+
+    /**
+     * @brief Creates an empty endpoint for late initialisation.
+     */
+    static auto make_uninitialised() -> endpoint { return endpoint{}; }
+
+    auto to_string() const -> std::string { return ip().to_string() + std::to_string(port()); }
```
jbaldwin: I'm wondering if this should this have a `:` between the two parts?

jbaldwin (changes requested):

jbaldwin:
Ok cool, thanks for checking!

Edit: it can be installed via the cmake variable but isn't on but default since git pre commit hooks are weird to force on everyone

jbaldwin (approved):

PR #450, opened 2026-03-05
coro::thread_pool_ws (work stealing)

jbaldwin:
This is a proof of concept that would replace the normal coro::thread_pool if it works out, this uses a chase lev work stealing queue with an atomic list for the global queue when schedules happen outside a thread pool worker. It leverages thread_local uint32_t to lookup the current threads index for its owned internally local thread queue, if that isn't set then we know we're off a worker thread and need to enqueue on the global queue.

### DIFF
--- a/.githooks/readme-template.md
+++ b/.githooks/readme-template.md
@@ -608,9 +608,9 @@ If you require TLS, provide OpenSSL for the target ABI (static or shared) and se
 
 ### Requirements
     C++20 Compiler with coroutine support
-        g++ [11, 12, 13]
-        clang++ [16, 17]
-        MSVC Windows 2022 CL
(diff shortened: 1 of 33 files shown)


PyXiion:
Using CPM for dependencies is good, I wanted to suggest it sooner or later

Consider to migrate other dependencies too

jbaldwin:
> Using CPM for dependencies is good, I wanted to suggest it sooner or later
> 
> Consider to migrate other dependencies too

Yeah I think its a great idea for the submodules (yay).

It looks like for this change I need to edit this dependency for emscripten to get it to compile/work, so its going to just be embedded into the repo under details like I did for the `tl::expected`, but luckily its only one header file so no big deal.

PR #452, opened 2026-03-12
IOCP support (again)

PyXiion:
Well, there's a lot of work to do to do that.

Right now I need to write/rewrite these things:
 - [ ] `io_notifier_iocp`
 - [ ] `pipe_t`: rename it to `wake_source` or something like that, and implement it on Windows via `PostQueuedCompletionStatus` and custom completion keys
 - [ ] `poll_info` & `poll_op`
 - [ ] `timer_handle`: 
   1. `std::multimap`/`std::set` with timers, and calculate dynamic timeout for `GQCS(iocp, ..., timeout_ms)`.
    On new timer wake IOCP queue with `PQCS(..., TIMER_WAKE_KEY, nullptr)`.
      - Pros: one thread, no thread pool
      - Cons: heap usage, possibly bad performance with many timers, maybe precision loss
   2. `CreateThreadpoolTimer` + `PostQueuedCompletionStatus` from the callback
      - Pros: OS will manage it itself, high precision, simpler code
      - Cons: overhead for short timers, thread pool used
   3. `CreateWaitableTimer` + dedicated waiter thread
      - Pros: super duper simple
      - Cons: extra thread and overall bad


PyXiion:
I think I'll try to move all async operations into io_notifier to make the API consistent on both Windows & Unix

```cpp
class io_notifier_epoll
{
  // ...

    // renamed poll_info
    auto watch(async_operation& pi) -> bool;
    
    // ...

    // Operations
    auto write_some(net::socket &socket, std::span<const std::byte> buffer) -> coro::task<std::pair<net::io_status, ssize_t>>;
    auto read_some(net::socket &socket, std::span<std::byte> buffer) -> coro::task<std::pair<net::io_status, ssize_t>>;
    
    // File I/O may look like this in the future
    // auto write_some(coro::file &file, std::span<const std::byte> buffer) -> coro::task<std::pair<io_status, ssize_t>>;
};
```

PR #457, opened 2026-03-29
make -Wall, -Wextra and /W4 private to the project

### LINKED ISSUE #453: Setting project related warnings should probably be set to PRIVATE and not PUBLIC
Currently libcoro is setting a bunch of flags (all, extra) to public. To me humble knowledge these are project specific settings and should therefore be applied privately and not publicy.

I would like to propose to make the following change, from:
```cmake
    target_compile_options(${PROJECT_NAME} PUBLIC
        $<$<COMPILE_LANGUAGE:CXX>:-fexceptions>
        $<$<COMPILE_LANGUAGE:CXX>:-Wall>
        $<$<COMPILE_LANGUAGE:CXX>:-Wextra>
        $<$<COMPILE_LANGUAGE:CXX>:-pipe>
    )
```
to:
```cmake
    target_compile_options(${PROJECT_NAME} PUBLIC
        $<$<COMPILE_LANGUAGE:CXX>:-fexceptions>
        $<$<COMPILE_LANGUAGE:CXX>:-pipe>
    )
    target_compile_options(${PROJECT_NAME} PRIVATE
        $<$<COMPILE_LANGUAGE:CXX>:-Wall>
        $<$<COMPILE_LANGUAGE:CXX>:-Wextra>
    )
```

And yes, of course every project should _always_ use Wall and Wextra. But sometimes you don't, or a project has a very good reason of not wanting to use these.

daantimmer:
This PR sets the -Wall, -Wextra and /W4 flags scoped to just the libcoro project.
It also removes the duplicate flags in libcoro_test and removed flags -fcoroutines and -fconcepts (these are set by default when the language version is set to C++20)
- ref: https://gcc.gnu.org/onlinedocs/gcc/C_002b_002b-Dialect-Options.html#index-fcoroutines
- ref: https://gcc.gnu.org/onlinedocs/gcc/C_002b_002b-Dialect-Options.html#index-fconcepts

`-std=c++20` could be removed from libcoro_test because its always set through CMAKE_CXX_STANDARD
`-fexceptions` is removed because its set publicly in libcoro

Fixes #453

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -207,7 +207,7 @@ if(DEFINED EMSCRIPTEN)
     add_link_options(-sEXIT_RUNTIME)
 endif()
 
-if(${CMAKE_CXX_COMPILER_ID} MATCHES "GNU|Clang")
+if(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
     add_compile_options(
         $<$<COMPILE_LANGUAGE:CXX>:-Wall>
         $<$<COMPILE_LANGUAGE:CXX>:-Wextra>
@@ -259,7 +259,7 @@ if(LIBCORO_FEATURE_NETWORKING)
     endif()
 endif()
 
-if(${CMAKE_CXX_COMPILER_ID} MATCHES "GNU")
+if(CMAKE_CXX_COMPILER_ID MATCHES "GNU")
     if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS "10.2.0")
         message(FATAL_ERROR "g++ version ${CMAKE_CXX_COMPILER_VERSION} is unsupported, please upgrade to at least 10.2.0")
     endif()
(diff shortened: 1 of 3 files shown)


daantimmer:
So, I can also confirm from a 'consuming project' point of view that this is now properly set 'privately':

currently using my forked version here: https://github.com/philips-software/amp-cucumber-cpp-runner/pull/312

And the actual commit with that change (and what I could remove because of it): https://github.com/philips-software/amp-cucumber-cpp-runner/pull/312/commits/963455f1330991b415d3b252fdc4a9a786c9754c

jbaldwin:
I think the windows build is busted because GitHub added windows-2025 which windows-latest now points at. I think if you adjust the tank to be windows-2022 it should work again and we can get a new issue to get 2025 working.

jbaldwin (approved):

